PR #7, opened 2015-03-26
Document the use of UTF-8 by default

### LINKED ISSUE #4: Document character set assumption
The library implicitly assumes the PHP 'default_charset' option is correctly set for the data passed, as it is calling htmlspecialchars with a charset option, which will prompt conversions.
That's perfectly reasonably, but I think you should document this.

adrilo:
Fixes #4


boxcla:

PR #8, opened 2015-03-28
Add support for multiline strings

### LINKED ISSUE #6: Error: Shared string not found for index
While reading excel sheet we are having a weird issue. We get the excel file from the external system. It runs fine if run unedited but throws following exception when we edit and save the same file. Even if we just add extra space or delete a sheet, the file throws the same exception.

> Shared string not found for index: 1432
#0 /opt/lampp/htdocs/mr/app/Vendor/Box/Spout/Reader/XLSX.php(247): Box\Spout\Reader\Helper\XLSX\SharedStringsHelper->getStringAtIndex(1432)
1 /opt/lampp/htdocs/mr/app/Vendor/Box/Spout/Reader/AbstractReader.php(131): Box\Spout\Reader\XLSX->read()

It displays random value which is not even in the current sheet. Note that we tested just by running the sample code provided under "How to read a XLSX file?" in documentation.

Any idea what may be going wrong?

adrilo:
Fixes #6 

Escaped line feed characters in shared strings before processing them. This makes every string remain on one single line and therefore allow fast retrieval
Replaced usages of "\n" by PHP_EOL
Added test for multiline strings


boxcla:

PR #11, opened 2015-04-02
Guess Excel cell type basing on array cell value type

gdaszuta:
Hello, 

writing to Excel files creates some problem on larger integers - this patch is guessing proper cell type basing on data, while currently used is_numeric() was causing writing large integers into output file disregarding variable type.

This patch can in some cases break compatibility for some scripts - perhaps it should be wrapped in conditional expression based on some parameter, allowing to revert to current behavior, yet I'm really bad at inventing parameter names.

Regards,
Grzegorz


adrilo:
Fixed in #18

PR #18, opened 2015-04-15
Better guess the cell type based on its value

adrilo:
Follow-up for #11 

Fixed the bugs caused by empty() and is_numeric(): 0 was skipped, "10" was considered as a number
Added tests


boxcla:

PR #21, opened 2015-04-16
Fix empty shared string bug

adrilo:
Replaced !$sharedString by $sharedString === null to avoid the case when $sharedString = ''


boxcla:

PR #29, opened 2015-04-29
Expose a Sheet object on Reader::XLSX::nextSheet()

### LINKED ISSUE #28: Get current worksheet's name when reading a XLSX file
Currently, $reader->nextSheet() does not return anything. It would be useful to have it return the current sheet and therefore expose its name.

adrilo:
Fixes #28 

Added Sheet class for the XLSX reader that exposes basic sheet info, such as name or ID.
When retrieving the sheet data XML, added extra XML parsing to retrieve sheet data.
Added test


boxcla:

PR #43, opened 2015-05-29
Add support for files containing formulas

### LINKED ISSUE #41: Can only read few columns
I try to read a large excel files (more than 30 columns with column range formulas) and 11 thousands rows. This library can only read 6 columns.

Any clue why ?

adrilo:
Fixes #41 

Formulas will be skipped on reading.
The result of the formulas will be kept though.


boxcla:

PR #44, opened 2015-05-29
Bump to version 1.0.9

adrilo:


boxcla:

stof:
the version field should actually be omitted entirely from the composer.json, relying on the git metadata instead. Setting it can only create issues for VCS repos:

- it is ignored entirely for branches (the version is set based on the branch name, and then branch aliases are applied)
- for tags, it must match the version guessed based on the tag name, otherwise the tag is considered as invalid and rejected by composer.

This means that the only effect of specifying the version explicitly is to allow mistakes in the release process making releases invalid: https://getcomposer.org/doc/04-schema.md#version

Specifying the version explicitly should be done only for package metadata specified manually (the ``package`` repository type)

adrilo:
Thanks for the tip @stof ! I removed it here: #46

PR #46, opened 2015-06-03
Remove version from composer.json

adrilo:
As mentionned here, version in composer.json can be omitted and only managed through git versioning: https://github.com/box/spout/pull/44#issuecomment-107912203


boxcla:

PR #48, opened 2015-06-03
Add support for more cell types

### LINKED ISSUE #47: Trying to get property of non-object: /vendor/box/spout/src/Spout/Reader/XLSX.php:227
Hi,

I get the following error:
 Trying to get property of non-object: /vendor/box/spout/src/Spout/Reader/XLSX.php:227
reading the following Excel with the following code:
```
$data = array();
        $reader = \Box\Spout\Reader\ReaderFactory::create(\Box\Spout\Common\Type::XLSX);
        $reader->open($file);
        $counter = 0;
        $header = array();
        while ($reader->hasNextSheet()) {
            $reader->nextSheet();
            while ($reader->hasNextRow()) {
                $counter++;
                if ($counter == 1) {
                    $header = $reader->nextRow();
                    continue;
                }
                $data[] = array_combine($header, $reader->nextRow());
            }
        }
        $reader->close();
        return $data;
```
2015-06-03 14_47_32-microsoft excel - template xlsx

adrilo:
Fixes #47 

Added proper support for booleans, dates, numbers, errors.
Added unescaping of the read string.
Fixed a bug when cells did not have any values => now returns empty string.


boxcla:

src/Spout/Reader/XLSX.php
```diff
@@ -239,6 +224,58 @@ protected function read()
...
+     * @return string|int|float|bool|null The value associated with the cell (null when the cell has an error)
+     */
+    protected function getCellValue($node, $escaper)
+    {
+        $cellValue = '';
+
+        // Default cell type is "n"
+        $cellType = $node->getAttribute('t') ?: 'n';
+
+        if ($cellType === 'inlineStr') {
```
Lewiscowles1986: Any way this could be further separated into calling two functions, one for inlineStr, and one for other cell types?

Lewiscowles1986: re-factor into switch-case with function per row-format

PR #49, opened 2015-06-11
Added option for SharedStringsHelper to turn off the file cache

Bob4ever:


boxcla:
Hi @Bob4ever, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

Bob4ever:
CLA signed

boxcla:
Verified that @Bob4ever has just signed the CLA. Thanks, and we look forward to your contribution.

src/Spout/Reader/Helper/XLSX/SharedStringsHelper.php
```diff
@@ -34,6 +34,9 @@ class SharedStringsHelper
     /** Value to use to escape the line feed character ("\n") */
     const ESCAPED_LINE_FEED_CHARACTER = '_x000A_';
 
+    /** This will increase your memory usage but can improve your execution time */
+    static public $KEEP_ALL_STRINGS_IN_MEMORY = false;
```
adrilo: this should be a protected variable with a public setter

src/Spout/Reader/Helper/XLSX/SharedStringsHelper.php
```diff
@@ -141,11 +144,19 @@ public function extractSharedStrings()
 
             $unescapedTextValue = $escaper->unescape($textValue);
 
-            // The shared string retrieval logic expects each cell data to be on one line only
-            // Encoding the line feed character allows to preserve this assumption
-            $lineFeedEncodedTextValue = $this->escapeLineFeed($unescapedTextValue);
-
-            $this->writeSharedStringToTempFile($lineFeedEncodedTextValue, $sharedStringIndex);
+            if (self::$KEEP_ALL_STRINGS_IN_MEMORY) {
+                if (!Isset($this->inMemoryContents)) {
```
adrilo: Isset => isset. I think you don't even need this check. PHP will accept you setting the array value directly

src/Spout/Reader/Helper/XLSX/SharedStringsHelper.php
```diff
@@ -263,25 +274,32 @@ protected function getSharedStringTempFilePath($sharedStringIndex)
...
-        $tempFilePath = $this->getSharedStringTempFilePath($sharedStringIndex);
-        $indexInFile = $sharedStringIndex % self::MAX_NUM_STRINGS_PER_TEMP_FILE;
-
-        if (!file_exists($tempFilePath)) {
-            throw new SharedStringNotFoundException("Shared string temp file not found: $tempFilePath ; for index: $sharedStringIndex");
-        }
+        $sharedString = null;
+    
+        if (self::$KEEP_ALL_STRINGS_IN_MEMORY) {
+            if (isset($this->inMemoryContents) && array_key_exists($sharedStringIndex, $this->inMemoryContents)) {
```
adrilo: isset => is_array to be safer

adrilo:
Thanks for the pull request! I left some comments. One thing we should not forget is to update the README to reflect these changes once we agreed on the final solution.

src/Spout/Reader/Helper/XLSX/SharedStringsHelper.php
Bob4ever: Mhh I found this problem in a document with ~130k shared strings. The problem is that the strings aren't in the 'right' order. In some rows the helper need to reread the cache files again and again and agian.

src/Spout/Reader/Helper/XLSX/SharedStringsHelper.php
Bob4ever: I think only without strict. But it should be cleaner to first define the array before filling it.

src/Spout/Reader/Helper/XLSX/SharedStringsHelper.php
Bob4ever: I changed this.

adrilo:
That's good stuff :)
Before approving, what do you think about pushing this idea a bit further: instead of manually setting the boolean, it could be automatically set based on the number of unique shared strings that will be read.
This way, if you need to read small and big reports, Spout will be smart enough to use the right method instead of being statically set. I think it would be a huge plus!!

src/Spout/Reader/Helper/XLSX/SharedStringsHelper.php
adrilo: I'm not a big fan of exposing this global method to support messed up files.
One of the goal of Spout is to read any file without going out of memory. This is something other libraries can't do and that I'd like to preserve. Adding an option to disable caching goes against this philosophy as it will make it possible to run out of memory (what if you had 500K strings instead of 130K?). In this case, I would argue that it's better to have a slow program than a one that crashes.

Does it make sense?

Bob4ever: Yes indeed. My problem is that this is VERY slow and grows exponential. In my case I tried to read the file with 130k shared strings and 60k rows. When I just read the first 1k rows spout needs ~8 seconds to read this, 5k rows = ~8sec but with 7k rows spout needs 30seconds. To read the complete file I needed more than 360seconds (I interrupted at that point maybe it would much longer maybe not...). I totally unterstand your problem with this 'fix'. This is just a fast way around in cost of memory. But the caching need to be done much more effective. In the worst case the shared strings helper read every cell a new cache file because the shared strings aren't really ordered by their position.

adrilo: The problem here is that there is no way to know how the strings are ordered unless you actually parse the whole sheet data file. And of course, you don't want to parse the whole file twice...
So given that you can't come up with a smart way to group strings or keep some strings in memory, what compromise would you choose?

My answer to that is the following: for optimization reason, the XLSX writer in Spout does not use shared strings as they should be. A duplicate string should only have one entry in the sharedStrings.xml file and be referenced twice in the sheet data file. Instead, and because we don't want to keep in memory which strings have already been processed, every string gets assigned a new index. Not size efficient but memory efficient!
This means that shared strings are ordered by appearance, making the current caching mechanism efficient.

Because that's the only scenario that I can control, and that I can't optimize for other scenarios, I decided to optimize the caching for this one.

PR #55, opened 2015-06-30
Coverage Added

Lewiscowles1986:
Coverage compatible with coveralls.io added, you would need to sign up free at coveralls.io, then add the repository and Travis-CI will automagically push the report. FYI: not sure if you know, but you have excellent coverage >95%

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ script: phpunit
 
 before_script:
   - composer self-update
-  - composer install --dev --optimize-autoloader --prefer-source
+  - composer install --dev --prefer-source
   - mkdir -p build/logs
 
 after_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,8 @@ before_script:
   - composer self-update
   - composer install --dev --prefer-source
   - mkdir -p build/logs
-  - phpunit --coverage-clover build/logs/clover.xml
 
-script: phpunit
+script: phpunit --coverage-clover build/logs/clover.xml
 
 after_script:
   - php vendor/bin/coveralls -v
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,12 @@ php:
   - nightly
   - hhvm
 
-install:
-  - composer self-update
-  - composer install --prefer-source
-
 script: phpunit
+
+before_script:
+  - composer self-update
+  - composer install --dev --optimize-autoloader --prefer-source
+  - mkdir -p build/logs
+
+after_script:
+  - php vendor/bin/coveralls -v
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ before_script:
   - composer self-update
   - composer install --dev --prefer-source
   - mkdir -p build/logs
+  - phpunit --coverage-clover build/logs/clover.xml
 
 script: phpunit
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,12 @@ php:
   - nightly
   - hhvm
 
-script: phpunit
-
 before_script:
   - composer self-update
   - composer install --dev --prefer-source
   - mkdir -p build/logs
+
+script: phpunit
 
 after_script:
   - php vendor/bin/coveralls -v
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -15,10 +15,17 @@
     </testsuites>
 
     <logging>
-        <log type="coverage-html" target="tests/coverage/"/>
+        <log type="coverage-clover" target="build/logs/clover.xml"/>
+        <log type="coverage-php" target="build/logs/coverage.cov"/>
+        <log type="coverage-text" target="build/logs/coverage.txt" showUncoveredFiles="false"/>
+        <log type="coverage-html" target="test/reports/coverage/" showUncoveredFiles="true"/>
+        <log type="junit" target="test/reports/junit.xml" logIncompleteSkipped="false"/>
     </logging>
 
     <filter>
+        <blacklist>
+            <directory suffix=".php">vendor/</directory>
+        </blacklist>
         <whitelist>
             <directory suffix=".php">src/</directory>
         </whitelist>
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -14,14 +14,6 @@
         </testsuite>
     </testsuites>
 
-    <logging>
-        <log type="coverage-clover" target="build/logs/clover.xml"/>
-        <log type="coverage-php" target="build/logs/coverage.cov"/>
-        <log type="coverage-text" target="build/logs/coverage.txt" showUncoveredFiles="false"/>
-        <log type="coverage-html" target="test/reports/coverage/" showUncoveredFiles="true"/>
-        <log type="junit" target="test/reports/junit.xml" logIncompleteSkipped="false"/>
-    </logging>
-
     <filter>
         <blacklist>
             <directory suffix=".php">vendor/</directory>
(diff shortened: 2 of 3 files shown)


boxcla:
Hi @Lewiscowles1986, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

Lewiscowles1986:
"CLA Signed". Thanks! :+1:

boxcla:
Verified that @Lewiscowles1986 has just signed the CLA. Thanks, and we look forward to your contribution.

phpunit.xml
```diff
@@ -15,10 +15,17 @@
     </testsuites>
 
     <logging>
-        <log type="coverage-html" target="tests/coverage/"/>
+        <log type="coverage-clover" target="build/logs/clover.xml"/>
+        <log type="coverage-php" target="build/logs/coverage.cov"/>
+        <log type="coverage-text" target="build/logs/coverage.txt" showUncoveredFiles="false"/>
+        <log type="coverage-html" target="test/reports/coverage/" showUncoveredFiles="true"/>
```
adrilo: why is this target test/reports while others are build/logs?

phpunit.xml
```diff
@@ -15,10 +15,17 @@
     </testsuites>
 
     <logging>
-        <log type="coverage-html" target="tests/coverage/"/>
+        <log type="coverage-clover" target="build/logs/clover.xml"/>
+        <log type="coverage-php" target="build/logs/coverage.cov"/>
+        <log type="coverage-text" target="build/logs/coverage.txt" showUncoveredFiles="false"/>
```
adrilo: the build directory should be added to .gitignore (if kept).

phpunit.xml
```diff
@@ -15,10 +15,17 @@
     </testsuites>
 
     <logging>
-        <log type="coverage-html" target="tests/coverage/"/>
+        <log type="coverage-clover" target="build/logs/clover.xml"/>
```
adrilo: do we need to generate all those reports? Or is only one enough?

adrilo: I found this article: http://mnapoli.fr/code-coverage-reports-with-github-travis-and/
It seems like we can get rid of the logging part alltogether (since it's annoying to have the reports generated everytime the unit tests are run) and only update the travis.yml file.
It seems like a better solution. What do you think?

Lewiscowles1986:
Ha gonna make myself look an idiot here... It's just what I have always done to get coveralls working, like the coveralls php dev-dep I added. I know it comes through, which is generally my main concern, but tbh I have less than 1% clue about the config of coveralls, it's one of those, it works and I know what coveralls itself is good for, so I don't worry so much about the config as long as it works :blush:

Lewiscowles1986:
also you dont add build to the gitignore, as it only exists on the build-server on travis. I suppose if the tests were run locally it might be a good idea, but, I am quite sure the best way to run the tests it to fetch a clean repo from github per-test, so the build folder should not exist

.travis.yml
```diff
@@ -7,8 +7,12 @@ php:
   - nightly
   - hhvm
 
-install:
+script: phpunit
+
+before_script:
   - composer self-update
-  - composer install --prefer-source
+  - composer install --dev --optimize-autoloader --prefer-source
```
adrilo: --optimize-autoloader is not really useful here as it's not a production environment. Generating the autoloader will become slower and the optimization gain is not enough to be justified here

.travis.yml
```diff
@@ -7,8 +7,12 @@ php:
   - nightly
   - hhvm
 
-install:
+script: phpunit
+
+before_script:
```
adrilo: can you move this section before "script" so that it goes in a chronological order: "before_script" - "script" - "after_script"

Lewiscowles1986:
if it makes you happy I will :+1: :smiley_cat:

adrilo:
BTW if you need to reply to a specific comment, you should click on the "Files changed" tab and reply from there directly (instead of using the global discussion) :smile:

phpunit.xml
Lewiscowles1986: I don't know if it will work, lol

phpunit.xml
Lewiscowles1986: the build folder is not added or existing in repo, so I think this is doing more than we have to. The idea is that we pull from GitHub and test clean every time

phpunit.xml
adrilo: I'm pretty sure it will :) By adding these lines in phpunit.xml, you make the configuration static. This config can also be defined at runtime using similar option:

`phpunit --coverage-clover build/logs/clover.xml` is the same as `<log type="coverage-clover" target="build/logs/clover.xml"/>`

phpunit.xml
Lewiscowles1986: Just how I have always done this, not sure what the effects of changing are, I didn't get mine from a blog, a fellow dev showed me their setup and it worked, unlike reading some articles, so I stuck with it. Not ideal, but the config setup is one of the least important parts of this as it does not affect the test outcome, or the repository

Lewiscowles1986:
:smile: I honestly feel like you might be trolling a little now lol

adrilo:
I was just trying to prevent the discussion from being too scattered. Some people are not too familiar with Github and it can quickly become really hard to follow a discussion... It looks like you know what you are doing, no offense then!! :)

phpunit.xml
adrilo: Right! That does not mean it's not important to make it right :)
Actually, if you make the change, we'll see immediately if it works as I enabled coveralls for the repo already. If Github shows the coverall report, then we're good!

Lewiscowles1986:
No sorry if it came across like that, It shows all comments here, so I was trying to add chronology, anyway, thanks for your time and input, it really is appreciated! :]

Lewiscowles1986:
right looks like most of the changes broke things :hushed: 
https://coveralls.io/builds/2939953 worked
since, not 1 build has been success...

.travis.yml
```diff
@@ -7,8 +7,12 @@ php:
   - nightly
   - hhvm
 
-install:
+before_script:
   - composer self-update
-  - composer install --prefer-source
+  - composer install --dev --prefer-source
+  - mkdir -p build/logs
```
adrilo: Being super picky here: I read the documentation on the php-coveralls website: https://github.com/satooshi/php-coveralls#travis-ci

Can you update this file to match what they have?

```
before_script:
 - wget http://getcomposer.org/composer.phar
 - php composer.phar install --dev --no-interaction

script:
 - mkdir -p build/logs
 - phpunit --coverage-clover build/logs/clover.xml

after_script:
 - php vendor/bin/coveralls -v
```

adrilo:
Regarding your last comment, I guess you can squash all your commits into one, as the final changes are pretty simple. It will make it less confusing in the git history :)

.travis.yml
Lewiscowles1986: you know composer is installed, and we are just updating it and telling it to get deps right?

I'll alter the script if it has to be done, but it being on the coveralls website, like the other website, which the recommendations of has lead to broken travis builds without changing any php source; can cause issues, which is why I was resistant to change before... Things that were working are no longer working.

adrilo: didn't the issue come from the double phpunit call?

Lewiscowles1986: No, if you look the build failed, and double PHP unit call was what I thought the issue was. Strangely it builds without issue in my repo, so not sure what is wrong

adrilo: https://coveralls.io/builds/2942765 it seems good to me. ??

Lewiscowles1986:
closing, reopening, then re-submitting as new ;)

PR #58, opened 2015-07-01
Prevent entity loading when reading XML

### LINKED ISSUE #57: Issue when building (intermittent)
Box\Spout\Reader\XLSXTest::testReadShouldBeProtectedAgainstBillionLaughsAttack

Even if I build master, this seems to fail to assert for PHP5.6 only... How did TravisCI pass the build, if this always fails?

adrilo:
Fixes #57 

Added LIBXML_NOENT option when reading a XML file
libxml_disable_entity_loader(true) cannot be used because it disables the use of XMLReader::open()... see https://bugs.php.net/bug.php?id=62577

### DIFF
--- a/src/Spout/Reader/Helper/XLSX/SharedStringsHelper.php
+++ b/src/Spout/Reader/Helper/XLSX/SharedStringsHelper.php
@@ -112,7 +112,7 @@ class SharedStringsHelper
         $escaper = new \Box\Spout\Common\Escaper\XLSX();
 
         $sharedStringsFilePath = $this->getSharedStringsFilePath();
-        if ($xmlReader->open($sharedStringsFilePath, null, LIBXML_NONET) === false) {
+        if ($xmlReader->open($sharedStringsFilePath, null, LIBXML_NOENT|LIBXML_NONET) === false) {
             throw new IOException('Could not open "' . self::SHARED_STRINGS_XML_FILE_PATH . '".');
         }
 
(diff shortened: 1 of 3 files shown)


boxcla:

Lewiscowles1986:
:+1:

Eitz:
Hi @adrilo,

I would like to comment on this old PR as it appears that the code is contradicting what the title says. I blame this on totally confusing constant naming (LIBXML_**NO**ENT) and the lack of documentation on it. It actually does the opposite its name says: when you use it, it **turns on** Entity interpretation/substitution.

With that in mind, I question: are we sure this library is not vulnerable to XXE Attacks?

You can check for more information here:

1. https://stackoverflow.com/questions/38807506/what-does-libxml-noent-do-and-why-isnt-it-called-libxml-ent
2. https://security.stackexchange.com/questions/133906/is-php-loadxml-vulnerable-to-xxe-attack-and-to-other-attacks-is-there-a-list
3. https://www.owasp.org/index.php/XML_External_Entity_%28XXE%29_Processing

adrilo:
Hi @Eitz, 

Thanks for your feedback! This is a very old PR indeed. If you look at the code now, it does not use the NOENT anymore (https://github.com/box/spout/blob/0f0bf648024040d21c6bd9930299265e8fc42550/src/Spout/Reader/Wrapper/XMLReader.php#L32). We also realized that was a mistake :)

It's actually stated clearly in the doc now: https://github.com/box/spout/blob/0f0bf648024040d21c6bd9930299265e8fc42550/tests/Spout/Reader/ODS/ReaderTest.php#L281

Eitz:
Awesome @adrilo, thanks for the good work.

Issue #65, opened 2015-07-07
Option to use SplFixedArray for shared strings, instead of file-based cache.


kewlar:
I have a proposal [to have an option] to make Spout faster by using SplFixedArray instead of files for storing shared strings. At the moment, switching shared-string cache files is one of the biggest performance hogs. On small files, it's not that big a deal. On big files it becomes painful.

I slapped together a quick spike to check how storing shared strings in files vs. array vs. SPL array compare in an XLSX reading test. Here are the benchmarks on my machine:

**Test1.xlsx:** 1.28 MB XLSX file, 9884 rows, 1.04 MB sharedStrings.xml

|           | memory_get_peak_usage(), MB | reader->open() time, s | Total read time, s |
| :-------- | --------------------------: | ---------------------: | -----------------: |
| Files     | 1.43                        | 7.03                   | 288.89             |
| Array     | 4.91                        | 3.54                   | 22.55              |
| SPL array | 2.97                        | 3.51                   | 22.79              |

**Test2.xlsx:** 79 MB XLSX file, 859988 rows, 53.5 MB sharedStrings.xml

|           | memory_get_peak_usage(), MB | reader->open() time, s | Total read time, s |
| :-------- | --------------------------: | ---------------------: | -----------------: |
| Files     | 1.59                        | 290.11                 | 20945.36           |
| Array     | 224.58                      | 162.93                 | 1300.42            |
| SPL array | 137.32                      | 174.85                 | 1262.81            |

I suggest to have an option in Spout to use in-memory SharedStringsHelper, when needed, which would then use SplFixedArray for storing them.

You can see my changes to the code at https://github.com/kewlar/spout/commit/bc72e2c34e9bb18c02a2be169c3d63917694ea1f, if you like. It was meant only for benchmarking the cases, but you can see that the SplFixedArray implementation for shared strings takes little enough code. I didn't figure out how to elegantly tell ReaderFactory (or XLSX) which shared strings helper strategy to use, though.

I understand that one of the goals of Spout was to use as little memory as possible, and I think the way it was accomplished was (and still is) ingenious, and I understand you may want to keep it this way. Still, in some use cases it may be preferable to sacrifice some memory for greater processing speed.

What do you think?

adrilo:
Your idea is similar to what was done here: #49. I'm definitely for such a change but I believe that we can do even better :)

[...]

PR #67, opened 2015-07-13
Caching strategies

adrilo:
Fixes #49 and #65 

Implementation of a new caching strategy for the shared strings: in-memory. This should speed up the reading of small XLSX files!
To keep the library simple, the choice of the best strategy has been delegated to an algorithm that takes into account the number of unique shared strings as well as the memory available.


boxcla:

PR #73, opened 2015-07-21
Moved readers to iterators

### LINKED ISSUE #50: Inconsistent interface
### Problem

As it stands the return type of `ReaderFactory::create()` method is `XLSX` while it can return both `CSV` and `XLSX`. Additionally `XLSX` defines methods (`hasNextSheet()` and `nextSheet()`) that are not present in neither `CSV` nor `ReaderInterface` 

### Proposed Solution

1. Change the return type of `ReaderFactory::create()` to `ReaderInterface`
2. Add the missing methods to the `ReaderInterface`
3. Implement the new interface methods in the `CSV` class as empty methods.

The method implementation (3\.) could be made at the `AbstractReader` level, but since `csv` is likely to be the only supported format that has no sheets it makes more sense to keep the sheet-less implementation at the `CSV` level.

As far as I am aware

### Usage Example

Here's an example function that seamlessly process CSV and single-sheet XLSX files:
```php
function processSpreadsheet(\Box\Spout\Reader\ReaderInterface $reader, $filePath)
{
    $reader->open($filePath);
    if (method_exists($reader, 'hasNextSheet')) { // don't want to check for concrete type in case more formats are implemented
        $reader->nextSheet();
    }

    processRows($reader);

    $reader->close();
}
```

With the described changes the above function can be simplified to this:
```php
function processSpreadsheet(\Box\Spout\Reader\ReaderInterface $reader, $filePath)
{
    $reader->open($filePath);
    $reader->nextSheet(); // XLSX reader will select the first sheet, CSV reader will do nothing

    processRows($reader);

    $reader->close();
}
```

### LINKED ISSUE #63: Iterator support
*Branching of from discussion in #50*

As @adrilo suggested, it could look something like this:

```php
    foreach ($sheet->getRowIterator() as $row) {
       // do stuff
    }
}
```

Your suggestion was to replace the `hasNext` and `next` methods entirely, but I am wondering if it would be feasible to have both to maintain backwards compatibility. What should happen if the old methods are called during iteration via iterators?

adrilo:
Fixes #50 and #63 

Instead of the hasNext() / next() syntax, readers now implements the PHP iterator pattern. It allows readers to be used with a foreach() loop.

All readers now share the same structure (CSV is treated as having exactly one sheet):
- one concrete Reader
- one SheetIterator, exposed by the Reader
- one or more Sheets, returned at every iteration
- one RowIterator, exposed by the Sheet

Introducing the concept of sheets for CSV may be kind of confusing but it makes Spout way more consistent. Also, this confusion may be resolved by creating a wrapper around the readers if needed.


boxcla:

src/Spout/Reader/XLSX/RowIterator.php
```diff
@@ -0,0 +1,356 @@
...
+     * @return void
+     * @throws \Box\Spout\Common\Exception\IOException If the sheet data XML cannot be read
+     */
+    public function rewind()
+    {
+        $this->xmlReader->close();
+
+        $sheetDataFilePath = 'zip://' . $this->filePath . '#' . $this->sheetDataXMLFilePath;
+        if ($this->xmlReader->open($sheetDataFilePath, null, LIBXML_NONET) === false) {
+            throw new IOException('Could not open "' . $this->sheetDataXMLFilePath . '".');
```
Lewiscowles1986: Needs test for exception, or exception removed

Lewiscowles1986:
#### Pros

* Increased coverage
* I like the iterator syntax as it is more purposeful, and shorter
* Looks like a significant portion of code went to the graveyard
* Better division of code

#### Cons

* Changed namespaces (agreed `Internal/XLSX` is less clear than `XLSX/Internal` ), will this break anything? Should converting be added? Release bump needed?
* Not all exceptions tested, and at least one exception removed for EOF
* Does each type of cell need to become it's own type of entity?

src/Spout/Reader/XLSX/RowIterator.php
adrilo: I'll add tests for all the thrown exceptions as well as the rewind() functions. I need to make this that if you run foreach(...) twice, you read the file twice.

I removed a bunch of exceptions on purpose. With the iterator syntax, I don't think I should worry about people calling rewind(), next(), current() functions directly (because they are implicitely called in the foreach). It is a different situation than the previous one where hasNext() and next() were explicitely called and could therefore be called in the wrong order...

adrilo:
Thanks for the comments @Lewiscowles1986! Totally agree with the pros :) You can add that the library is now more consistent (same interface for CSV and XLSX).

Regarding the cons:

> Changed namespaces (agreed Internal/XLSX is less clear than XLSX/Internal ), will this break anything? Should converting be added? Release bump needed?

Yes, this change is not backwards compatible... As mentioned here: https://github.com/box/spout/issues/63#issuecomment-123177748, I tried many things to have a smooth transition between the 2 iteration methods, but it was very painful and created a lot of mess. So I decided it was not worth it... And you're right, iterators will be introduced with Spout 2.0 :)

> Not all exceptions tested, and at least one exception removed for EOF

I removed a bunch of exceptions on purpose. With the iterator syntax, I don't think I should worry about people calling rewind(), next(), current() functions directly (because they are implicitely called in the foreach). It is a different situation than the previous one where hasNext() and next() were explicitely called and could therefore be called in the wrong order...
For the non tested exceptions, I can add some tests. With the new architecture, it became harder to test things at the row iterator level (since the public methods are 4 levels higher, on the reader). But I'll add the needed tests.

> Does each type of cell need to become it's own type of entity?

IMO this would add some unnecesary complexity. What we can do is move the cell value extraction into its own helper, to make the code cleaner. But since the returned value of the reader is an array of values representing a row, actually returning an array of cells would only make sense if each type of cell would expose some specific functionality.

PR #76, opened 2015-07-28
Csv multiple encodings

### LINKED ISSUE #72: How to set input encoding for Reader?
Hi @adrilo!

How I can set input encoding 'CP1251' for reading file?

For instance, in PHPExcel that can be achieved with the following:
```php
$objReader = new PHPExcel_Reader_CSV();
$objReader->setInputEncoding('CP1251');
```

adrilo:
Fixes #72 

Added support for multiple encodings with CSV files (through $reader->setEncoding($encoding)).
It requires either iconv or php-intl extensions to be installed. Hopefully this won't be a problem!


boxcla:

PR #77, opened 2015-07-29
Fix CSV reader when last line is empty

adrilo:
If the last line was empty, it would create an infinite loop...


boxcla:

PR #78, opened 2015-07-29
Fix XMLReader open() overriding

adrilo:
This is to avoid a warning in PHP7 (and also because that's how it should be!)


boxcla:

PR #83, opened 2015-08-14
Set wrap text style when multiline string encountered

### LINKED ISSUE #10: Set cell style to "wrap text" when contains multi-line string
Currently Excel does not display multi-line strings correctly. On Mac, new line characters are replaced with a space while on Windows, it is ignored. Internally though, Excel has the right data (the new line can be seen in the formula bar).

In order to fix the display, the "wrap text" attribute needs to be turned on for these cells. Spout should autodetect new lines and set this attribute automatically.

adrilo:
Fixes #10
If a cell contains a multiline string, "wrap text" style option should automatically be set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,7 +159,7 @@ $style = (new StyleBuilder())
            ->setFontSize(15)
            ->setShouldWrapText()
            ->build();
-           
+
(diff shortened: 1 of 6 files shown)


boxcla:

PR #87, opened 2015-08-19
Use Travis cache

adrilo:
This is to avoid re-downloading all dependencies for every build.


boxcla:

PR #91, opened 2015-08-21
Detection of invalid sheet name

### LINKED ISSUE #90: Generated XLSX shows error when opened in MS Excel
There seems to be an error in workbook.xml, as I always get a message dialog when opening generated files in MS Excel. Files can be opened but they need to be repaired. Are you aware of the issue?

adrilo:
Fixes #90 

Based on Excel requirements:
 - it should not be blank
 - it should not exceed 31 characters
 - it should not contain these characters: \ / ? * [ or ]
 - it should be unique

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,6 +263,15 @@ $sheet = $writer->getCurrentSheet();
 $sheetName = $sheet->setName('My custom name');
 ``` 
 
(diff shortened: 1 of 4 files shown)


boxcla:

PR #92, opened 2015-08-21
Detection of invalid sheet name - continued

adrilo:
Invalid names can also be triggered by:
- character ":"
- single quote at the beginning of the name
- single quote at the end of the name

Introduced a StringHelper, wrapping multibyte strings functions


boxcla:

PR #96, opened 2015-08-24
Throw if XLSX Writer configured after being opened

### LINKED ISSUE #95: Throw exception when setShouldUseInlineStrings is called after openWriter
setTempFolder(), setShouldUseInlineStrings() and setShouldCreateNewSheetsAutomatically() must be called before opening the writer, otherwise they have no effect.
Throwing an exception if this is done after would make it explicit

adrilo:
Fixes #95


boxcla:

PR #99, opened 2015-08-29
ODS writer

adrilo:
Added ODS writer
Refactored XLSX writer to abstract some pieces into an abstract multi-sheets writer
Created an abstract style helper
Moved shared components around


boxcla:

PR #102, opened 2015-08-31
Improve ODS Writer

adrilo:
Remove num-columns-repeated and num-rows-repeated as it does not seem to be required (LibreOffice does not add them). This greatly simplifies the writer and the XML output.
Added some optional attributes to help LibreOffice with cell values caching ("calcext").


boxcla:

PR #103, opened 2015-08-31
Use number-columns-repeated in ODS writer

adrilo:
The number-columns-repeated usage may reduce the size of the outputted XML file by merging repeated values together.


boxcla:

PR #106, opened 2015-09-01
ODS Reader

### LINKED ISSUE #45: ODS output
Hi I have a repo for fast-efficient writing to ODS,and I was wondering if it is a file-format you are interested to have as part of spout?

adrilo:
Fixes #45 

Spout can now read ODS files.
It's on par with the XLSX reader. The only difference is that the row iterator cannot be rewound.
It supports the different output formats from LibreOffice and Excel, skipping extra rows/cells if needed.


boxcla:

PR #124, opened 2015-10-09
Support for style per column for XLSX and ODS

### LINKED ISSUE #123: Support for column-based styles for XLSX and ODS
Instead of row based styles it would be nice to have style for each column (which is inserted by row).

JWolters:
resolves #123

### DIFF
--- a/src/Spout/Writer/AbstractWriter.php
+++ b/src/Spout/Writer/AbstractWriter.php
@@ -209,11 +209,15 @@ abstract class AbstractWriter implements WriterInterface
      */
     public function addRowWithStyle(array $dataRow, $style)
     {
-        if (!$style instanceof Style\Style) {
-            throw new InvalidArgumentException('The "$style" argument must be a Style instance and cannot be NULL.');
+        $styles = is_array($style) ? $style : [$style];
+
+        foreach ($styles as $style) {
+            if (!$style instanceof Style\Style) {
+                throw new InvalidArgumentException('The "$style" argument must be a Style instance and cannot be NULL.');
+            }
         }
 
-        $this->setRowStyle($style);
+        $this->setRowStyle($styles);
         $this->addRow($dataRow);
         $this->resetRowStyleToDefault();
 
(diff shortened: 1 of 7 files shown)


boxcla:
Verified that @JWolters has signed the CLA. Thanks for the pull request!

adrilo:
Hey @JWolters, thanks for the pull request!

Few general comments:
- you can squash all the commits together (or have one for the code, one for the tests). It makes it easier to understand the history
- tests should pass

In general, I'm not a fan of having functions that can accept multiple types of argument (here, Style or Style[]). It is confusing for the caller and usually becomes hard to maintain and test.
So being able to support multiple styles per row should be done - imo - differently. To be honest, I'm still not sure how this can be nicely achieved, with the current architecture (where Spout's interface only cares about rows, not cells). If you have any thoughts on this, please let me know!

JWolters:
Hi Adrilo,

I started the styles per row for my own benefit: I wanted to have a column formatted like a date so people can order by that column. I didn’t want to change to much code that’s why I have used the multiple types of argument.

For our purpose the first commit was what we needed, but not for the various tests. The only successfully tested commit was with changes in the test code I did not like, and then I had to do other work.

My actual problem is that what is called styles in Spout is only some font formatting but not the style as can be used in XLSX (or ODS). I want to use and Excel date (days since 1900-01-01) and show it in a yyyy-mm-dd format.

tomikulmala:
Would it be possible to add column background color and font color to this too?

RastoStric:
It is pity that this thing got stuck here :(

adrilo:
I believe that the right way to solve this is to make Spout more granular. Currently, the more granular element is the row (represented as an array of values). If we go one step further, we can introduce a per cell behavior. This means, you could style each cell separately.

It involves a lot of changes but is, I think, a reasonable way to address a lot of the issues that were filed (including this one). Looking forward to v3 :)

adrilo:
Stale... closing as this will be done in v3

PR #131, opened 2015-10-23
Better date support

### LINKED ISSUE #81: No "date" type when reading XSLX
When reading a XSLX file the output types are not always what I would expect.

Observations From limited testing with some xslx files from Excel and exported document from Google Sheets give met the same result:

A cell with a Date or DateTime value appears to be converted to a FLOAT (number of days since 1899-12-30 as per Lotus 123 tradition) rather than a PHP `DateTime` value.

This is unexpected as `Spout/Reader/XLSX/RowIterator.php` explicitly defines cell type `CELL_TYPE_DATE` to convert Date values.

Similarly, a numeric value resulting from an evaluated expression like `=12+13` is converted to a string value `"25"` rather than an `Int` or `Float` value of `25`

Not sure if this is a bug or a limitation related to XSLX variants?

adrilo:
Fixes #81 

Although Excel has a Date type, older Excel versions use numeric values to store dates.
The value represents the number of days since Jan 1st, 1900.
The only way to tell if the value is a number or a date is to look at the styles.xml and check if the cell has date formatting.


boxcla:

PR #138, opened 2015-11-05
Increase entropy of uniqid

adrilo:
This is to avoid conflicts if two folders are being created at the exact same time.


boxcla:

PR #143, opened 2015-11-11
[XLSX] Skip empty cells on write

### LINKED ISSUE #140: Too big xlsx
http://i.imgur.com/5PQx25U.png
csv file size 14 MB
must remove empty fields
http://i.imgur.com/PwSDQFL.png

/vendor/box/spout/src/Spout/Writer/XLSX/Internal/Worksheet.php
http://i.imgur.com/zSJq5la.png

after remove empty fields
http://i.imgur.com/kSG1Mze.png
but very slowly opened
compared to xlsx created from csv with help excel

adrilo:
Fixes #140 

Since cells are referenced by their coordinates (A2, B4...), it is not necessary to write empty cells.
This will reduce the final size of the generated XML and therefore XLSX file.


boxcla:

PR #153, opened 2015-12-06
Improve ZIP interface for better mime detection

adrilo:
The ZipHelper interface is now more generic and allow single files to be added.
It supports adding uncompressed files (for PHP7+), which is required to have the mime detection magic work with ODS files.
Also fixed a few issues with the created ODS file (thanks to https://odf-validator.rhcloud.com/)


boxcla:

PR #165, opened 2016-01-08
Support XLSX that are defined in random order

### LINKED ISSUE #163: getSheetIterator() returns sheets not in the right order
(First: we love box/spout, it works really well and we use it a lot. Thanks!)

We use box/spout for a 'product import' where we only use the first sheet. This works well in most cases but now we experienced something weird. Looks like some versions of Excel declare their sheets in [Content_Types].xml in a different order then they are displayed in Excel. Excel opens them in the 'right' way, but box/spout in the order they appear in the xml.

When we open such a 'bad' Excel in the latest Excel for Mac, it opens normally. When we save a copy of it from that Excel, the sheets in the xml are in the right order and now box/spout returns the sheets in the right order.

Apparently, Excel uses another location where they store the order of sheets. **Is it possible to change the order of sheets in box/spout to exactly as they appear in Excel?**

screen_shot_2015-12-24_at_14_08_15

We now 'fixed' it by looping through the sheets until we find one with at least 1 row in it. But it looks like a bug to us. Maybe there is a way to fix this in box/spout the right way?

adrilo:
Fixes #163 

Some software generate [Content_Types].xml file with sheets definition in random order.
Instead of having the first sheet (id = 1) defined first, it may be defined in 3rd position.
Therefore, to read the file in the correct order, sheets order need to be fixed.


boxcla:

PR #167, opened 2016-01-14
Fix CellValueFormatter for numeric values

### LINKED ISSUE #159: Invalid handling of long numbers in XLSX
When library loads long number (f.e. 865640023012945), it is converted to float (in \Box\Spout\Reader\XLSLX\CelllValueFormatter - formatNumericCellValue), and, because of this, number losts its last digits.

In some cases it's completely unacceptable behavior, maybe it's better to return raw string in that case (when is_numeric($val) == true, is_int($val) == false, is_float($val) == false) ?

As I can see, ODS has the same problem, it tries to convert long number to float too (\Box\Spout\Reader\XLSLX\CelllValueFormatter - formatFloatCellValue).

adrilo:
Fixes #159 

The value passed into the format() function is coming from an XML file and has never been coerced.
Therefore, when checking is_int($value), the check always returns false - because it's a string.
Changing the check fixes the issue and Spout now correctly parses large numbers.

### DIFF
--- a/src/Spout/Reader/ODS/Helper/CellValueFormatter.php
+++ b/src/Spout/Reader/ODS/Helper/CellValueFormatter.php
@@ -47,7 +47,6 @@ class CellValueFormatter
 
     /**
      * Returns the (unescaped) correctly marshalled, cell value associated to the given XML node.
-     * @TODO Add other types !!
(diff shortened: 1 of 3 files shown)


boxcla:

PR #175, opened 2016-02-15
Fixing the boolean bug.

IlyaBakhlin:
Before this coomit, if you add a `false` value, the cell in the Excel document is empty, because generated string is the following:

```
<c r="B1" s="0" t="b"><v></v></c>
```

But, with this fix, the string is the following:

```
<c r="B1" s="0" t="b"><v>0</v></c>
```

So, the `false` value is generated propperly and it's not being confused with the `null` value generation.

### DIFF
--- a/src/Spout/Writer/XLSX/Internal/Worksheet.php
+++ b/src/Spout/Writer/XLSX/Internal/Worksheet.php
@@ -146,7 +146,11 @@ EOD;
                     $cellXML .= ' t="s"><v>' . $sharedStringId . '</v></c>';
                 }
             } else if (CellHelper::isBoolean($cellValue)) {
-                $cellXML .= ' t="b"><v>' . $cellValue . '</v></c>';
+                if ($cellValue === true) {
+                    $cellXML .= ' t="b"><v>1</v></c>';
+                } else {
+                    $cellXML .= ' t="b"><v>0</v></c>';
+                }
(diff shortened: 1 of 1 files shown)


boxcla:
Hi @IlyaBakhlin, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

IlyaBakhlin:
CLA signed

boxcla:
Verified that @IlyaBakhlin has just signed the CLA. Thanks, and we look forward to your contribution.

src/Spout/Writer/XLSX/Internal/Worksheet.php
```diff
@@ -146,7 +146,11 @@ public function addRow($dataRow, $style)
                     $cellXML .= ' t="s"><v>' . $sharedStringId . '</v></c>';
                 }
             } else if (CellHelper::isBoolean($cellValue)) {
-                $cellXML .= ' t="b"><v>' . $cellValue . '</v></c>';
+                if ($cellValue === true) {
```
adrilo: this can be simplified using `intval`:
`$cellXML .= ' t="b"><v>' . intval($cellValue) . '</v></c>';`

IlyaBakhlin:
Done. :v: :sweat_smile:

PR #177, opened 2016-02-16
PhpStorm autocompletion compatibility.

IlyaBakhlin:
If the documentation comment specifies that a `WriterInterface` is returned, PhpStorm can't recognize that there's a `setTempFolder` method. But with this fix, it recognizes all the available methods `Writer` classes have. And, in my opinion, the documentation comment is more correct.

### DIFF
--- a/src/Spout/Writer/WriterFactory.php
+++ b/src/Spout/Writer/WriterFactory.php
@@ -20,7 +20,7 @@ class WriterFactory
      *
      * @api
      * @param  string $writerType Type of the writer to instantiate
-     * @return WriterInterface
+     * @return CSV\Writer | XLSX\Writer | ODS\Writer
      * @throws \Box\Spout\Common\Exception\UnsupportedTypeException
      */
     public static function create($writerType)


boxcla:
Verified that @IlyaBakhlin has signed the CLA. Thanks for the pull request!

src/Spout/Writer/WriterFactory.php
```diff
@@ -20,7 +20,7 @@ class WriterFactory
      *
      * @api
      * @param  string $writerType Type of the writer to instantiate
-     * @return WriterInterface
+     * @return CSV\Writer | XLSX\Writer | ODS\Writer
```
adrilo: There are pros and cons to both methods. I actually had the enumeration of all possible types before but changed it to use the interface instead.

Pros of listing all types:
 - better autocomplete
 - more descriptive

Cons:
 - too much autocomplete (if you have a CSV writer, you should only get autocomplete for the CSV writer methods, not the XLSX ones)
 - harder to maintain (if you add a new type, you have to remember to add it here and in its reader counterpart)

I believe that the cons still outweigh the pros. I'd rather have less autocomplete but safer code (because all of the autocompleted methods are available). And when I need autocomplete, I add a PHPDoc hint on the variable itself. PHPStorm does support this very well.

For instance:
```php
/** @var \Box\Spout\...\CSV\Writer $writer */
$writer = WriterFactory::create(Type::CSV);

/** @var \Box\Spout\...\XLSX\Writer $writer */
$writer = WriterFactory::create(Type::XLSX);
```

adrilo: What do you think?

IlyaBakhlin:
As the `WriterFactory` class is, as its name indicates, it's a factroy, the comments should specify the classes it returns. If there's a lot of autocompletion information, in my opinion, it's a bad design problem.

Although, it's your library, and I'm only providing changed I'd like to have if it'd my library.

adrilo:
The PHPDoc is used to specify a contract on the returned value. It does not necessarily need to be comprehensive though. Returning an interface and not a list of concrete classes is pretty common.

For a correct autocomplete, returning an interface or a list of classes won't change the fact that you need to either cast or give your IDE a hint about the created writer. Otherwise you end up either with methods that are not available (if you have a list of class) or missing methods (if you return an interface).

If none of the options actually solve the problem, I'd rather keep the return type as an interface, for the reasons explained above

PR #188, opened 2016-03-19
Custom stream wrapper support

### LINKED ISSUE #176: Stream wrapper support (at least for CSV)
As mentioned in #132 with the addition of a `realpath()` in the `open()` it is not possible to open a file with a streamwrapper. Among other stream wrappers Amazon S3 uses stream wrappers to bring support for standard file operations using `s3:\\[filepath]`. I'm trying to read a CSV file from S3, but the only issue is the `realpath()` (I tested it without the `realpath()`). I understand it's much harder to find a solution for `XMLReader()`, but that class is not used for CSV files. Is there some way to skip the realpath check for streamwrappers (for example with config setting or param)?

adrilo:
Fixes #176

Added support for custom stream wrappers, such as "fly" or "s3".
Support is determined per reader.

### DIFF
--- a/src/Spout/Common/Helper/GlobalFunctionsHelper.php
+++ b/src/Spout/Common/Helper/GlobalFunctionsHelper.php
@@ -281,6 +281,17 @@ class GlobalFunctionsHelper
     }
 
     /**
+     * Wrapper around global function stream_get_wrappers()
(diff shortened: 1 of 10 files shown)


boxcla:

Issue #184, opened 2016-03-13
ODS Reader yields different results


madflow:
Hi there!

When playing with the ```Reader``` part of the library I came across the artifact, that the ODS reader treats input differently than its peers XLSX and CSV.

1. 0 (Zeros) are treated as missing values - in a Spreadsheet these are perfectly valid values (PHP empty() in use...) I would argue, that even "empty" values are perfectly valid and should yield empty arrays in the row iterator (Like in CSV and XLSX).

**Example**:
```
use Box\Spout\Writer\WriterFactory;
use Box\Spout\Reader\ReaderFactory;
use Box\Spout\Common\Type;


$data = [
    ['A','B','C'],
    [1, 2, 3],
    [0, 0, 0]
];

// Write Data
$writerXlsx = WriterFactory::create(Type::XLSX);
$writerXlsx->openToFile('test.xlsx');
$writerXlsx->addRows($data);
$writerXlsx->close();

$writerOds = WriterFactory::create(Type::ODS);
$writerOds->openToFile('test.ods');
$writerOds->addRows($data);
$writerOds->close();

$writerCsv = WriterFactory::create(Type::CSV);
$writerCsv->openToFile('test.csv');
$writerCsv->addRows($data);
$writerCsv->close();

// Read files and dump the result
$result = [];
$reader = ReaderFactory::create(Type::XLSX);

$reader->open('test.xlsx');
foreach ($reader->getSheetIterator() as $sheet) {
    foreach ($sheet->getRowIterator() as $row) {
        $result['xlsx'][] = $row;
    }
}
$reader->close();

$reader = ReaderFactory::create(Type::ODS);
$reader->open('test.ods');
foreach ($reader->getSheetIterator() as $sheet) {
    foreach ($sheet->getRowIterator() as $row) {
        $result['ods'][] = $row;
    }
}
$reader->close();

$reader = ReaderFactory::create(Type::CSV);
$reader->open('test.csv');
foreach ($reader->getSheetIterator() as $sheet) {
    foreach ($sheet->getRowIterator() as $row) {
        $result['csv'][] = $row;
    }
}
$reader->close();

print_r($result);

Array
(
    [xlsx] => Array
        (
            [0] => Array
                (
                    [0] => A
                    [1] => B
                    [2] => C
                )

            [1] => Array
                (
                    [0] => 1
                    [1] => 2
                    [2] => 3
                )

            [2] => Array
                (
                    [0] => 0
                    [1] => 0
                    [2] => 0
                )

[...]

PR #189, opened 2016-03-19
Fixes for #184

madflow:
* 0 (Zeros) were treated as missing values - in a Spreadsheet these are perfectly valid values. empty() has been replaced with a helper method that honours zeros as valid values.

* The calculated values to show in a row were computed differently between ODS and XLSX/CSV. This was due to the attempt to hinder the ODS reader to create <=16384 empty cells when the file has been created by Excel (2010 ? ). This was done by convention - when the last cell (node) in a row was "empty" and the cell should be repeated (number-columns-repeated), than no values where pushed to the list. This convention had side effects on perfectly valid empty values, that should be repeated. I replaced the convention with a vendor specific ```hack``` - this feels unnecessary - but something's got to give when you are dealing with different vendors.

I worked with what I saw in the code - there can be side effects I do not know of.


boxcla:

src/Spout/Reader/ODS/RowIterator.php
```diff
@@ -20,6 +20,7 @@ class RowIterator implements IteratorInterface
     const XML_NODE_TABLE = 'table:table';
     const XML_NODE_ROW = 'table:table-row';
     const XML_NODE_CELL = 'table:table-cell';
+    const MAX_COLUMNS_EXCEL= 16384;
```
adrilo: add a space between const and =

src/Spout/Reader/ODS/RowIterator.php
```diff
@@ -127,22 +128,23 @@ public function next()
...
+                    $isEmptyRow = ($numCellsRead <= 1 && $this->isEmptyCellValue($cellValue));
                     if ($isEmptyRow) {
                         // skip empty rows
                         $this->next();
                         return;
                     }
 
-                    // Only add value if the last read cell is not empty or does not need to repeat cell values.
+                    // Only add the value if the last read cell is not a trailing empty cell repeater in Excel.
+                    // The current count of read columns is determined my counting the the values in $rowData.
```
adrilo: ... BY counting THE... (my => by, only one THE)

adrilo:
Minor comments, otherwise looking good! Thanks @madflow

adrilo:
Thanks!!

PR #199, opened 2016-04-12
Check file handle is valid before manipulating it

adrilo:
This fixes issues when something went wrong on reader/writer init and the developer wants to close the reader/writer.
The file handle may not be defined so we need to add a check for it, before actually using it.


boxcla:

Issue #193, opened 2016-03-24
Can't read cell formatted as "time" in XLSX


bartoffw:
I've found a problem when reading cell values that are formatted as "time". In such case Excel just puts a float value (between 0.0 and 1.0) in the cell that precisely identifies the time of the day.
After some investigation I figured out that the cell is correctly identified as "datetime" format. However, as the value is below 1.0, it gets ignored [here](https://github.com/box/spout/blob/v2.4.3/src/Spout/Reader/XLSX/Helper/CellValueFormatter.php#L192) and returns null.

I have some quickfix for this, which basically treats such value as a "time" value and converts it correctly, but, as I'm not an XLSX format expert, I assume that it won't be correct for all cases.
I can provide a test XLSX file for this if needed.

adrilo:
Hi @bartoffw,

Thanks for filing this issue and starting the investigation!
It seems like we need to be more precise about the cell types. Dates and times have different requirements so instead of identifying a cell as "datetime", we should be able to tell if a cell is "date", "time" or "date+time". This way, we can handle the 0 - 1 edge case properly.

Feel free to submit a pull request if you can !

joefreire:
i'm with same problem, i need get value as time and spout return all time value as null
how i can solve that?

bartoffw:
@joefreire, @adrilo I've just submitted an initial fix for this, for those who need instant solution. As soon as I have some more time I'll think it over and see how the code may be improved.

joefreire:
thank you for you help, works for me

PR #202, opened 2016-04-20
Xlsx time

bartoffw:
Quick approach to fixing the #193 issue. It still requires some review and making sure the cell content will be properly recognized in all situations.

### DIFF
--- a/src/Spout/Reader/XLSX/Helper/CellValueFormatter.php
+++ b/src/Spout/Reader/XLSX/Helper/CellValueFormatter.php
@@ -189,7 +189,17 @@ class CellValueFormatter
 
         // The value 1.0 represents 1900-01-01. Numbers below 1.0 are not valid Excel dates.
         if ($nodeValue < 1.0) {
-            return null;
+            // treat this value as time-only cell
+            $time = round($nodeValue * 86400);
+            $hours = round($time / 3600);
+            $minutes = round($time / 60) - ($hours * 60);
+            $seconds = round($time) - ($hours * 3600) - ($minutes * 60);
+
+            // the date part is not important here
+            $dateObj = date_create('1-Jan-2000');
+            $dateObj->setTime($hours,$minutes,$seconds);
+
+            return $dateObj; 
         }
 
         // Do not use any unix timestamps for calculation to prevent


boxcla:
Hi @bartoffw, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

bartoffw:
CLA signed

boxcla:
Verified that @bartoffw has just signed the CLA. Thanks, and we look forward to your contribution.

src/Spout/Reader/XLSX/Helper/CellValueFormatter.php
```diff
@@ -189,7 +189,17 @@ protected function formatExcelTimestampValue($nodeValue)
 
         // The value 1.0 represents 1900-01-01. Numbers below 1.0 are not valid Excel dates.
         if ($nodeValue < 1.0) {
-            return null;
+            // treat this value as time-only cell
+            $time = round($nodeValue * 86400);
```
adrilo: I'd create 2 functions to handle both cases (time and date). Also, can you move the check for < 1.0 (that can be replaced by < 1 I believe) before the leap year fix above? the leap year is only relevant in the case of dates.

So something like this:

```php
if ($nodeValue < 1) {
  return $this->formatExcelTimestampAsTimeValue($nodeValue);
} else {
  return $this->formatExcelTimestampAsDateValue($nodeValue);
}
```

src/Spout/Reader/XLSX/Helper/CellValueFormatter.php
```diff
@@ -189,7 +189,17 @@ protected function formatExcelTimestampValue($nodeValue)
 
         // The value 1.0 represents 1900-01-01. Numbers below 1.0 are not valid Excel dates.
         if ($nodeValue < 1.0) {
-            return null;
+            // treat this value as time-only cell
```
adrilo: I'd create 2 functions to handle both cases (time and date). Also, can you move the check for < 1.0 (that can be replaced by < 1 I believe) before the leap year fix above? the leap year is only relevant in the case of dates.

So something like this:

```php
if ($nodeValue < 1) {
  return $this->formatExcelTimestampAsTimeValue($nodeValue);
} else {
  return $this->formatExcelTimestampAsDateValue($nodeValue);
}
```

src/Spout/Reader/XLSX/Helper/CellValueFormatter.php
```diff
@@ -189,7 +189,17 @@ protected function formatExcelTimestampValue($nodeValue)
...
+            $time = round($nodeValue * 86400);
+            $hours = round($time / 3600);
```
adrilo: `round` should be replaced by `floor` because you don't want to round up

src/Spout/Reader/XLSX/Helper/CellValueFormatter.php
```diff
@@ -189,7 +189,17 @@ protected function formatExcelTimestampValue($nodeValue)
...
+            $minutes = round($time / 60) - ($hours * 60);
```
adrilo: use `floor` here too

src/Spout/Reader/XLSX/Helper/CellValueFormatter.php
```diff
@@ -189,7 +189,17 @@ protected function formatExcelTimestampValue($nodeValue)
...
+            $seconds = round($time) - ($hours * 3600) - ($minutes * 60);
```
adrilo: no need to round `$time` as it was already done previously

src/Spout/Reader/XLSX/Helper/CellValueFormatter.php
```diff
@@ -189,7 +189,17 @@ protected function formatExcelTimestampValue($nodeValue)
...
+
+            // the date part is not important here
+            $dateObj = date_create('1-Jan-2000');
```
adrilo: Please use OO style:

```php
$dateObj = (new DateTime('1900-01-01'));
```

Also, let's use the default Excel date (Jan 1st, 1900) as the default date

src/Spout/Reader/XLSX/Helper/CellValueFormatter.php
adrilo: can you replace all numbers with constants defined at the top of the file?

NUM_SECONDS_IN_ONE_DAY, NUM_SECONDS_IN_ONE_HOUR, NUM_SECONDS_IN_ONE_MINUTE

src/Spout/Reader/XLSX/Helper/CellValueFormatter.php
```diff
@@ -189,7 +189,17 @@ protected function formatExcelTimestampValue($nodeValue)
...
+            $dateObj->setTime($hours,$minutes,$seconds);
+
+            return $dateObj;
```
adrilo: Can you please add a test for your change?

adrilo:
Thanks for the pull request! I left some comments but overall going in the right direction :)

adrilo:
Copied the changes into #224 and added tests. Thanks @bartoffw !
Closing the Pull Request as it was no longer active.

Issue #205, opened 2016-04-27
Numbers stored as string


arisdario:
Is there a way to avoid this?, for me is fine to get some columns with data stored as text, but in some case where i have only values like: "10.00" or "19.99" i don't want to store this kind of values as text, i need to select each cell and convert to value manually and if have 20k rows is really painful ! sometimes work with all selected sometimes some value break this selection !. Thx

adrilo:
Unfortunately, it's not currently possible. Spout inserts all values with the "General" type.
I'll add this to the todo list for v3 though.

arisdario:
Thanks !

jaminion:
I thought it might be worth mentioning that when you set your array elements that you pass to addRow() or addRows(), you can explicitly cast them to have them set as a numeric value in the cell instead of a string, at least with XLSX. In \Box\Spout\Writer\XLSX\Internal\Worksheet, the addRow() function uses CellHelper::isNumeric() to determine if the value is numeric, which in turn checks the return value from gettype() for that value.

Assuming you are pushing data into your spreadsheet from an array or object:

$data = array($obj->myStringValue(), (double)$obj->myDecimalValue(), (int)$obj->myIntValue());
$writer->addRow($data);

This is currently working for me.

@adrilo, your work on this project is much appreciated.

arisdario:
@adrilo, thanks for your hint. I will test this version and also i will test the performance and speed of writing a sheet whit 40k values. and i will post the rrsults. Thanks

Issue #182, opened 2016-03-09
Writing into cells with defined format


vika8713:
There would be very nice to have an opportunity to manage cell format while writing into. Now every entry has a general format (at least in .xlsx files).

adrilo:
Do you mean controlling the style for each cell separately instead of each row?

madflow:
> Do you mean controlling the style for each cell separately instead of each row?

@adrilo I believe that is what @vika8713  is asking for. This is basically covered  in #124 I think. I have a use case for this, where I want to create a "heatmap": a carpet of values (with background color) in a spreadsheet.

In addition to #124 I could think of lighweight "Cell Objects" that contain the cell value, the value type, styling, cell comments, other metadata - that could be passed to Writer::addRows().

TonisOrmisson:
I would love to have the opportunity to be able to format each cell individually. 
In terms of using spreadheet as an analysis tool it is very useful to be able to visually highlight cells to visually make it easier to identify important pieces in a matrix. I would have a lot of use for a tool that could do cell level formatting and would not run into memory problems over ridiculously small files (like phpExcel and others do).

TonisOrmisson:
Since the Styling currently is rather limited I'll specify that I would really love to be able to add to current styling: background colour, borders, cell comments. 
This would allow do create visually presentable tables.

BTW. I love this code. I have just gone through 4-5 different Excel readers/writers over past weeks that all seem to run into memory issues very quickly. This project seems exactly what I was looking for. 

If only had also cell formatting it could be used also in analysis type of tasks instead of raw import/export :)

madflow:
@adrilo Can we colaborate on how to proceed with the cell styling/formatting? My first instinct was to make the concept of cells a "first class citizen" and skip the "column styling" approach like documented here https://github.com/box/spout/pull/209#issuecomment-226816522. However I do not want to take another shot at this and see if it fits in a PR. 

Have you generated any ideas on how to proceed? Are there any "junior" tasks that can be adressed on the way to 3.0?

adrilo:
Hey @madflow, thanks for following-up on this issue. Sorry it took so long to get back to you!

I was thinking of the same approach, where instead of only accepting scalars or DateTime, the `addRow()` can also accept a `Cell` object. The `Cell` object holds data related to the cell that will be written (value and style for now).
I agree with the "skipping column styling", as it can be accomplished by cell styling (not as convenient for this use case, but doable).

[...]

Issue #151, opened 2015-12-05
Writing date


adrilo:
It is currently not possible to pass a DateTime object to a writer, while readers do support dates.

dexcell:
+1

iamquan:
:+1:

bevhost:
I made a workaround for this, since in Excel dates are just a number with a date format for presentation, if you output the right number you can format the column as a date once you load it into Excel and the correct date shows.
I use this  (strtotime($datestr)/86400)+25569.4167
strtotime converts the date string into a unix timestamp which is the number of seconds since 1970.  Excel dates are the number of days since 1900.  There is 86400 seconds in a day and 1970 is 25569 days from 1900.  I use .4167 because I'm in Australia which is GMT+10 hours.

adrilo:
The issue was created as a solution for the "appending data to an existing spreadsheet that contains dates" problem. To append data, one needs to copy the contents of the existing spreadsheet, row by row, and write each row to another spreadsheet before adding new data. This works well in most cases but did not work when cells contain dates: the reader is able to read dates, returning a DateTime object which is not accepted by the writer when trying to copy the row. 

This was before the `setShouldFormatDates` was added though... Now with this option, dates will be returned as strings, which is supported by the writer. So even though the cell type will be incorrect, Excel is usually good enough to detect that it's a date. And visually, the old and new spreadsheet will look exactly the same.

If a developer wants to write a date in another context, then it will be his/her responsibility to format the date before passing it to Spout.

I believe that this approach is a good compromise to avoid adding a ton of complexity to the Writer code to support dates properly. Hence closing this issue.

PR #209, opened 2016-05-04
Initial commit to support Cell styling and number formats in styles

Rushleader:
Quick fix for issue #205, #182 and #151; Also added a feature being able register the style to the sheet without having to call addRowWithStyle (required to add it to the cell as a single style)

Edit: I know it's a dirty fix. Will update this to tidy it up and support default format by using constants but would like this to be in the master so it's available to everyone.

Edit 2: For now its only for XLSX, since I dont have OpenOffice available atm


boxcla:
Verified that @Rushleader has signed the CLA. Thanks for the pull request!

Rushleader:
Usage example:
- Defining a number format:
```php
$styleRedFloat = (new StyleBuilder())->setNumberFormat('[Red]0.00000')->build();
$writer->registerStyle($styleRedFloat); # If using it for a single cell
```

- Append to the cell/row
```php
$row = array(
    array('cell1', $styleRedFloat)
);
$writer->addRow($row);

// Or

$row = array('cell1', 'cell2');
$writer->addRowWithStyle($row, $styleRedFloat);
```

Rushleader:
In an unexpected turn of events the number formatter can actually parse dates as custom format aswell :| 

```php
$styleDate = (new StyleBuilder())->setNumberFormat('d-mmm-YY HH:mm:ss')->build();

// Where time is a unix timestamp
$row[] = array(25569 + (time() / 86400), $styleDate);
```

This will format it correctly since the unix timestamp is changed to a julian excel date

Rushleader:
@adrilo Could you take a look at this ? kinda sitting here doing nothing atm

rocketpastsix:
As users of this repo, we would love to see this feature merged in. It would help us immensely.

adrilo:
@Rushleader, I'm gonna be really busy this week but I'll try to take a look at it next week. It's been a while since you pushed this PR... thanks for your patience!

src/Spout/Writer/Style/Style.php
```diff
@@ -56,11 +56,27 @@ class Style
     /** @var bool Whether specific font properties should be applied */
     protected $shouldApplyFont = false;
 
+    protected $verticalAlignment = 'center';
```
adrilo: There are 2 different changes here: alignment and number format. Can you please split these changes into 2 branches? It will reduce the scope of each commit and make the review easier/faster. Thanks!

test.php
```diff
@@ -0,0 +1,113 @@
+<?php
```
adrilo: Tests should be written with PHPUnit (look at examples in the `tests` folder)

src/Spout/Writer/XLSX/Internal/Worksheet.php
```diff
@@ -133,10 +133,18 @@ public function addRow($dataRow, $style)
 
         $rowXML = '<row r="' . $rowIndex . '" spans="1:' . $numCells . '">';
 
-        foreach($dataRow as $cellValue) {
+        foreach($dataRow as $cell) {
+            if (is_array($cell)) {
```
adrilo: I'm not a fan of this, where a value can have multiple types.

adrilo:
@Rushleader, your PR changes many things: alignment, number format, per-cell styling. That's too many things at once to be reviewed and merged safely. It will need to split into multiple PRs.

Now regarding your actual changes, number format does need per-cell styling (number format at the row level is not really useful). I'm not a fan of the way you did per-cell styling (with multiple possible types (scalar/array) for the same variable). I believe that the cell granularity should be exposed explicitly, with meaningful APIs, etc. This will come with v3...

So what we can do is eventually merge your change for the alignment but wait until v3 to get the number formatting change merged.

src/Spout/Writer/XLSX/Internal/Worksheet.php
Rushleader: Why shouldn't a value be able to be generated with multiple types (I agree, would be nicer with a class instead of whats happening now). Since a value and its interaction within the sheet can be different depending on type, especially when running around with different formats

test.php
Rushleader: This was just a quick test file to test what the performance decrease/impact would be. Since the premise of being able to style cells would seem te be a lot heavier

Rushleader:
@adrilo Yea the alignment change came from a mate of mine that wanted to use the number_formatting. He kinda pushed it into the same branch =/ 

I personally think that decent number formatting is kind of a big deal in excel. That's why I was surprised in the first place when I couldn't find it in the code. So in my opinion V3 might be a bit late for that feature.

Touching on the multi feature PR, that's always gonna be there since these features are dependent on one another (required the per cell/column styling for both of them).

src/Spout/Writer/XLSX/Internal/Worksheet.php
adrilo: A parameter that can be of multiple types hides the complexity of the program. If you pass an number, it will do X but if you pass an array it will do Y. While it's fine to do different things, the expectation of what a function does should be clear. In this example, there should probably be 2 functions (one taking a number, one taking an array) with explicit names.

The problem is less the difference between a number and a string but more between a single element VS a collection. Those represent really different things and should be treated carefully.

adrilo:
Alignment can be done on a per-row basis for now and can be migrated to a per-cell basis whenever this is ready, right? Even though some users will need to only apply the alignment to some cells, it may help with some use cases. Still it's not top priority (left alignment is acceptable in 99% of the cases).

Also regarding number formatting, I agree that this is pretty important. Spout was initially created as a CSV replacement: this means no formatting at all! Formatting was added progressively and there still is a lot more than can be added to it. This must also be balanced with performance, as the more features you add, you more you degrade performance. So not all features are necessarily worth being added.

Rushleader:
Hmm, i agree with you on the last point 'Not all features are worth being added' but I think for this to grow we need to atleast support a way to format a single cell (which requires both these features). 

The performance issue you touch on is a valid one, but it'll take a while before the performance hit will be noticeable. As with features not all being worth to be added, same goes for arbitrary restrictions (i.e. a performance decrease for a feature set). I understand that's hard to keep balanced though.

Ill try and see if i can tidy this up and split this into separate PRs. One for cellstyling, one for number formatting and one for alignment settings. (I think it's easier to call this "Column Styling" in the future since that's what we're shooting for.) Would this be easier to accept and keep in check ?

p.s. If Spout is getting this much bigger, it might be a good idea to start using a git-flow like structure and make the master branch protected and working in release branch cycles. That way it's also easier for people to create their own branch of Spout. 
We've been using this structure for a while now and it makes it easy to read the changes.
[http://nvie.com/posts/a-successful-git-branching-model/](url)

adrilo:
I'd rather go with Cell styling and skip column styling all together. Column styling can be achieved with cell styling so we should try to implement the most specific case that encapsulates more generic cases.
Alignment or number formatting are not necessarily column only options and can be applied to cells independently.

Anyways, splitting up this PR into 3 would make things a lot easier to review and therefore faster to merge. I'm maintaining this project on my spare time so when I see a big commit, I need to have quite some time ahead of me to start looking at it. Not something I can do in few minutes.
Regarding cell styling, I have some strong opinions on that so it may be better for me to tackle it. Then you can build all your improvements on top of it. Or we can do it in parallel and reconcile the 2 at the end.

Thanks for the link. I was looking into it a few weeks ago, as I started thinking about v3. Instead of committing to master, all v3 commits should be merged into the v3 branch and once done, we can merge the v3 branch back to master. Nothing new here though, just need to set it up :)
The git-flow approach is really interesting too. I'm wondering if this is preventing some developers to submit pull requests though, as it's yet another process that needs to be followed.

Rushleader:
I think it would make developer commit sooner since the merging process for develop should be simpler and less restrictive. During the time it's in develop it'll also get airtime on other branches and get tested on a per use basis (which by default will cover more weak points then unit testing, not excluding those btw).

And on the cell styling part, strong opinions? Whats ur vision for that specific scenario since I kinda need it to format it to specific float and date notations without losing the value typing.

adrilo:
Staled. Closing

ic-hv:
Why not just merge these changes in ...? They are useful!

adrilo:
It's not that easy unfortunately. This PR did not have tests, and had a few other things that did not meet the quality requirements we have on this project.
Now I agree that the features it brought are great and will definitely be useful. By the way, cell styling was recently added on the develop branch. We don't have number formatting nor alignment but it's coming, for all file types (not just XLSX)

bs-thomas:
Hello there, I'm wondering if there are updates on this issue?  Thanks guys!

adrilo:
Some of the features have been merged into the 3.0 branch (yet to be released).

bs-thomas:
Thanks for the quick response.  Wow looking forward to it!

juanmiguel431:
Now exists the method setFormat(). I think the documentation should be update it with this.

Example:
$styleDate = (new StyleBuilder())->setFormat('dd-mm-YYYY')->build();

adrilo:
Included in #687

PR #211, opened 2016-05-06
Add background-color to styles

madflow:
This is loosely related to #182. This PR introduces setting a background color to the styles definition. 

* For xlsx and ods.
* Only "solid" backgrounds are supported. XLSX supports a whole array of ```patternType``` constants (http://www.officeopenxml.com/SSstyles.php, darkGrid, lightTrellis) - these are ignored.
* XLSX knows a bgColor element (The color to use for the background of the fill when the type is not solid.) and a fgColor element (The color to use for the the background in solid fills.) - these are just treated as the same. Only a sick, twisted minded MS Office Interop Developer knows why a background color in solid fills is a foreColor. (I am kidding of course).
* Some Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ $writer->setShouldAddBOM(false);
 
 #### Row styling
 
-It is possible to apply some formatting options to a row. Spout supports fonts, borders as well as alignment styles.
+It is possible to apply some formatting options to a row. Spout supports fonts, background, borders as well as alignment styles.
 
 ```php
 use Box\Spout\Common\Type;
@@ -146,6 +146,7 @@ $style = (new StyleBuilder())
            ->setFontSize(15)
            ->setFontColor(Color::BLUE)
            ->setShouldWrapText()
+           ->setBackgroundColor(Color::YELLOW)
            ->build();
 
 $writer = WriterFactory::create(Type::XLSX);
--- a/src/Spout/Writer/Style/Style.php
+++ b/src/Spout/Writer/Style/Style.php
@@ -71,6 +71,13 @@ class Style
      */
     protected $shouldApplyBorder = false;
 
+    /** @var string Background color */
+    protected $backgroundColor = null;
+
+    /** @var bool */
+    protected $hasSetBackgroundColor = false;
+
+
     /**
      * @return int|null
      */
@@ -280,6 +287,35 @@ class Style
     }
 
     /**
+     * Sets the background color
+     * @param $color ARGB color (@see Color)
+     * @return Style
+     */
+    public function setBackgroundColor($color)
+    {
+        $this->hasSetBackgroundColor = true;
+        $this->backgroundColor = $color;
+        return $this;
+    }
+
+    /**
+     * @return string
+     */
+    public function getBackgroundColor()
+    {
+        return $this->backgroundColor;
+    }
+
+    /**
+     *
+     * @return bool Whether the background color should be applied
+     */
+    public function shouldApplyBackgroundColor()
+    {
+        return $this->hasSetBackgroundColor;
+    }
+
+    /**
      * Serializes the style for future comparison with other styles.
      * The ID is excluded from the comparison, as we only care about
      * actual style properties.
@@ -340,6 +376,9 @@ class Style
         }
         if (!$this->getBorder() && $baseStyle->shouldApplyBorder()) {
             $mergedStyle->setBorder($baseStyle->getBorder());
+        }
+        if (!$this->hasSetBackgroundColor && $baseStyle->shouldApplyBackgroundColor()) {
+            $mergedStyle->setBackgroundColor($baseStyle->getBackgroundColor());
         }
 
         return $mergedStyle;
--- a/src/Spout/Writer/Style/StyleBuilder.php
+++ b/src/Spout/Writer/Style/StyleBuilder.php
@@ -134,6 +134,19 @@ class StyleBuilder
     }
 
     /**
+     *  Sets a background color
+     *
+     * @api
+     * @param string $color ARGB color (@see Color)
+     * @return StyleBuilder 
+     */
+    public function setBackgroundColor($color)
+    {
+        $this->style->setBackgroundColor($color);
+        return $this;
+    }
+
+    /**
      * Returns the configured style. The style is cached and can be reused.
      *
      * @api
--- a/src/Spout/Writer/XLSX/Helper/StyleHelper.php
+++ b/src/Spout/Writer/XLSX/Helper/StyleHelper.php
@@ -4,6 +4,7 @@ namespace Box\Spout\Writer\XLSX\Helper;
 
 use Box\Spout\Writer\Common\Helper\AbstractStyleHelper;
 use Box\Spout\Writer\Style\Color;
+use Box\Spout\Writer\Style\Style;
 
 /**
  * Class StyleHelper
@@ -13,6 +14,64 @@ use Box\Spout\Writer\Style\Color;
  */
 class StyleHelper extends AbstractStyleHelper
 {
+    /**
+     * @var array
+     */
+    protected $registeredFills = [];
+
+    /**
+     * @var array [STYLE_ID] => [FILL_ID] maps a style to a fill declaration
+     */
+    protected $styleIdToFillMappingTable = [];
+
+    /**
+     * Excel preserves two default fills with index 0 and 1
+     * Since Excel is the dominant vendor - we play along here
+     *
+     * @var int The fill index counter for custom fills.
+     */
+    protected $fillIndex = 2;
+
+    /**
+     * XLSX specific operations on the registered styles
+     *
+     * @param \Box\Spout\Writer\Style\Style $style
+     * @return \Box\Spout\Writer\Style\Style
+     */
+    public function registerStyle($style)
+    {
+        $registeredStyle = parent::registerStyle($style);
+        $this->registerFill($registeredStyle);
+        return $registeredStyle;
+    }
+
+    /**
+     * Register a fill definition
+     *
+     * @param \Box\Spout\Writer\Style\Style $style
+     */
+    protected function registerFill($style)
+    {
+        $styleId = $style->getId();
+
+        // Currently - only solid backgrounds are supported
+        // so $backgroundColor is a scalar value (RGB Color)
+        $backgroundColor = $style->getBackgroundColor();
+
+        // We need to track the already registered background definitions
+        if (isset($backgroundColor) && !isset($this->registeredFills[$backgroundColor])) {
+            $this->registeredFills[$backgroundColor] = $styleId;
+        }
+
+        if (!isset($this->styleIdToFillMappingTable[$styleId])) {
+            // The fillId maps a style to a fill declaration
+            // When there is no background color definition - we default to 0
+            $fillId = $backgroundColor !== null ? $this->fillIndex++ : 0;
+            $this->styleIdToFillMappingTable[$styleId] = $fillId;
+        }
+    }
+
+
     /**
      * Returns the content of the "styles.xml" file, given a list of styles.
      *
@@ -84,13 +143,29 @@ EOD;
      */
     protected function getFillsSectionContent()
     {
-        return <<<EOD
-<fills count="1">
-    <fill>
-        <patternFill patternType="none"/>
-    </fill>
-</fills>
-EOD;
+        // Excel reserves two default fills
+        $fillsCount = count($this->registeredFills) + 2;
+        $content = sprintf('<fills count="%d">', $fillsCount);
+
+        $content .= '<fill><patternFill patternType="none"/></fill>';
+        $content .= '<fill><patternFill patternType="gray125"/></fill>';
+
+        // The other fills are actually registered by setting a background color
+        foreach ($this->registeredFills as $styleId) {
+
+            /** @var Style $style */
+            $style = $this->styleIdToStyleMappingTable[$styleId];
+
+            $backgroundColor = $style->getBackgroundColor();
+            $content .= sprintf(
+                '<fill><patternFill patternType="solid"><fgColor rgb="%s"/></patternFill></fill>',
+                $backgroundColor
+            );
+        }
+
+        $content .= '</fills>';
+
+        return $content;
     }
 
     /**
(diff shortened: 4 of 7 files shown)


boxcla:

tegansnyder:
@madflow you just made my day with this comment:

> "Only a sick, twisted minded MS Office Interop Developer knows why a background color in solid fills is a foreColor"

src/Spout/Writer/Style/Style.php
```diff
@@ -61,6 +62,13 @@ class Style
     /** @var bool Whether the wrap text property was set */
     protected $hasSetWrapText = false;
 
+    /** @var string Background color */
+    protected $backgroundColor = null;
```
adrilo: can this be set to self::DEFAULT_BACKGROUND_COLOR instead of null?

src/Spout/Writer/Style/Style.php
```diff
@@ -244,6 +252,34 @@ public function shouldApplyFont()
     }
 
     /**
+     * Sets the background color
+     * @param $color
+     */
+    public function setBackgroundColor($color = self::DEFAULT_BACKGROUND_COLOR)
```
adrilo: Remove default value. It's better to force callers to be explicit

src/Spout/Writer/Style/Style.php
```diff
@@ -302,6 +338,9 @@ public function mergeWith($baseStyle)
         if (!$this->hasSetWrapText && $baseStyle->shouldWrapText()) {
             $mergedStyle->setShouldWrapText();
         }
+        if (!$this->hasSetBackgroundColor && $baseStyle->shouldApplyBackgroundColor()) {
```
adrilo: should you add a check && $baseStyle->getBackgroundColor() !== self::DEFAULT_BACKGROUND_COLOR?
No need to set the background color if it's the default one.

src/Spout/Writer/Style/StyleBuilder.php
```diff
@@ -122,6 +122,19 @@ public function setShouldWrapText()
     }
 
     /**
+     *  Sets a background color
+     *
+     * @api
+     * @param $color
+     * @return $this
+     */
+    public function setBackgroundColor($color = Style::DEFAULT_BACKGROUND_COLOR)
```
adrilo: remove default value

src/Spout/Writer/XLSX/Helper/StyleHelper.php
```diff
@@ -84,13 +84,22 @@ protected function getFontsSectionContent()
...
-<fills count="1">
-    <fill>
-        <patternFill patternType="none"/>
-    </fill>
-</fills>
-EOD;
+        $registeredStyles = $this->getRegisteredStyles();
+        $fillsCount = count($registeredStyles);
+        $content = sprintf('<fills count="%d">', $fillsCount);
+        foreach ($registeredStyles as $style) {
```
adrilo: [nitpick] new line before foreach

src/Spout/Writer/XLSX/Helper/StyleHelper.php
```diff
@@ -84,13 +84,22 @@ protected function getFontsSectionContent()
...
+            if ($style->shouldApplyBackgroundColor()) {
+                $content .= sprintf(
+                    '<fill><patternFill patternType="solid"><fgColor rgb="%s"/><bgColor rgb="%s"/></patternFill></fill>',
+                    $style->getBackgroundColor(),
```
adrilo: Make this a variable instead of repeating the call. Also, just checking: is it ok that both the fgcolor and bgcolor are the same?

src/Spout/Writer/XLSX/Helper/StyleHelper.php
```diff
@@ -84,13 +84,22 @@ protected function getFontsSectionContent()
...
+                    $style->getBackgroundColor()
+                );
+            } else {
+                $content .= '<fill><patternFill patternType="none" /></fill>';
+            }
+        }
+        $content .= '</fills>';
+        return $content;
```
adrilo: [nitpick] new line before return

src/Spout/Writer/XLSX/Helper/StyleHelper.php
```diff
@@ -84,13 +84,22 @@ protected function getFontsSectionContent()
...
-    <fill>
-        <patternFill patternType="none"/>
-    </fill>
-</fills>
-EOD;
+        $registeredStyles = $this->getRegisteredStyles();
+        $fillsCount = count($registeredStyles);
+        $content = sprintf('<fills count="%d">', $fillsCount);
+        foreach ($registeredStyles as $style) {
+            if ($style->shouldApplyBackgroundColor()) {
```
adrilo: should you add a check either here or in the function directly: && $style->getBackgroundColor() != DEFAULT_BG_COLOR?

src/Spout/Writer/XLSX/Helper/StyleHelper.php
```diff
@@ -139,7 +148,7 @@ protected function getCellXfsSectionContent()
         $content = '<cellXfs count="' . count($registeredStyles) . '">';
 
         foreach ($registeredStyles as $style) {
-            $content .= '<xf numFmtId="0" fontId="' . $style->getId() . '" fillId="0" borderId="0" xfId="0"';
+            $content .= '<xf numFmtId="0" fontId="' . $style->getId() . '" fillId="' . $style->getId() .  '" borderId="0" xfId="0"';
```
adrilo: If I understand correctly, there will be as many fills declaration as there are styles. So even if none of the styles uses a bg color, we'll still define N fills with `patternType="none"` and use them here. Is that correct?
If my understanding is correct, any way this can optimized by having only one "no fill" declaration?

tests/Spout/Writer/ODS/WriterWithStyleTest.php
```diff
@@ -110,11 +110,13 @@ public function testAddRowWithStyleShouldListAllUsedStylesInCreatedContentXmlFil
             ->setFontItalic()
             ->setFontUnderline()
             ->setFontStrikethrough()
+            ->setBackgroundColor()
```
adrilo: remove

tests/Spout/Writer/ODS/WriterWithStyleTest.php
```diff
@@ -236,6 +240,23 @@ public function testAddRowWithStyleShouldApplyWrapTextIfCellContainsNewLine()
         $this->assertFirstChildHasAttributeEquals('wrap', $customStyleElement, 'table-cell-properties', 'fo:wrap-option');
     }
 
+    public function testDefaultBackgroundStyle()
```
adrilo: Add PHPDoc

tests/Spout/Writer/XLSX/WriterWithStyleTest.php
```diff
@@ -232,6 +251,27 @@ public function testAddRowWithStyleShouldApplyWrapTextIfCellContainsNewLine()
         $this->assertFirstChildHasAttributeEquals('1', $xfElement, 'alignment', 'wrapText');
     }
 
+    public function testDefaultBackgroundColor()
```
adrilo: phpdoc

src/Spout/Writer/Style/Style.php
madflow: The default background color is no background color at all ("transparent" I guess). The default background color in my head was only meant to enable setting a background without a parameter.  (```Style::setBackgroundColor()```)  So a developer will have to explicitly "set" a background. Maybe we should drop the "default background" alltogether?

src/Spout/Writer/Style/StyleBuilder.php
madflow: See above. I agree with the "explicit is better" part.

src/Spout/Writer/XLSX/Helper/StyleHelper.php
madflow: See above.

src/Spout/Writer/XLSX/Helper/StyleHelper.php
madflow: I will have to digg into the Office Open XML specs and open some files :S

src/Spout/Writer/XLSX/Helper/StyleHelper.php
madflow: I got this from LibreOffice - They create the border definition for xlsx in that way. I actually have no Excel License :D so I can not check there.

src/Spout/Writer/Style/Style.php
adrilo: Enabling setting a background without parameter seems weird. If you don't set any background, it should use the default background but if you do call the `setBackgroundColor()` method, I think your intent is to override the default behavior. Hence you should explicitly provide the new background color to use.

So what I would do is remove the concept of default bg color. And you can keep the if check as it is now

src/Spout/Writer/XLSX/Helper/StyleHelper.php
adrilo: You can check with this test file: [bg-colors.xlsx](https://github.com/box/spout/files/315506/bg-colors.xlsx)

src/Spout/Writer/XLSX/Helper/StyleHelper.php
adrilo: Well, I guess the specs tell you that you can/should reuse common parts. Whether it's fills, borders, shared strings, it seems like this is what OOXML is recommending (and that makes sense).

README.md
```diff
@@ -125,7 +125,7 @@ The writer always generate CSV files encoded in UTF-8, with a BOM.
 
 #### Row styling
 
-It is possible to apply some formatting options to a row. Spout supports fonts as well as alignment styles.
+It is possible to apply some formatting options to a row. Spout supports fonts, background as well as alignment styles.
```
adrilo: you'll have to rebase and resolve the conflicts with borders

src/Spout/Writer/XLSX/Helper/StyleHelper.php
adrilo: I think we should find a way to reuse fills here. Creating separate styles can make the file become huge

madflow:
@adrilo Thanks for your support.  I rebased and merged the current upstream. 

I agree - we should use "best practice / reusing common parts" for the ```bgColor`` and ```fgColor``` definition and the ```fills```defininition.

This PR does turn out to be more than a quick afterwork hack - so I would appreciate any tips and pushes to the PR branch.

madflow:
It turns out, XLSX seems to preserve two "no background" definitions. I have a hard time finding the actual specification for this... No luck yet. 

But LibreOffice does it and openpyxl and probably everyone else.

```
    <fills count="2">
        <fill><patternFill patternType="none"/></fill>
        <fill><patternFill patternType="gray125"/></fill>
    </fills>
```

I'll amend the PR.

adrilo:
I don't think there's a spec for that. The spec will describe how `<fill>` works. Then each implementation does its own thing... Microsoft Excel also creates 2 `<fill>` so I think we should try to copy that :)

adrilo:
Regarding the reuse of the `fills`, here is some guidance:

In `AbstractStyleHelper.php`, when a style gets registered, we should extract the fill information and store it separately. What's stored can be a hash representing the fill data. You assign it an id that you set on the style.
Then every time you register a new style, you do the same thing: extract the fill data, hash it, check if it's already stored; if so, get the id and set it on the style; otherwise add the hash to the fills array with an associated id and set this id to the style.
Finally you write the fills content, you can go through the fills array instead of going through the styles.

This is high level but hopefully it can help. BTW, you can do this in another PR, as the changes involved may be tricky

madflow:
This one is getting very stale :/

I commited and pushed the latest changes. My first approach was to amend the AbstractHelper class - but when I was finished I decided that it should got into the XLSX StyleHelper...

Considering, that fonts and borders are both not re-using definitions in the XLSX world at the moment, this seemed correct. There should be more to come. 

- Wrote better tests
- Deleted bgColor from the XLSX definition. Openpyxl does not use it when just a "start" color is supplied, so I decided that they have already solved this problem.
- When we decide to introduce an "end" color - we can introduce it again.
- Hacked the two default styles by using a seperate counter for the fillIndex.

I am not perfectly happy how it looks at the moment - but it adds imediate business value. (background colors :dancer: )

src/Spout/Writer/XLSX/Helper/StyleHelper.php
```diff
@@ -84,13 +143,29 @@ protected function getFontsSectionContent()
...
+        $fillsCount = count($this->registeredFills) + 2;
+        $content = sprintf('<fills count="%d">', $fillsCount);
+
+        $content .= '<fill><patternFill patternType="none"/></fill>';
+        $content .= '<fill><patternFill patternType="gray125"/></fill>';
+
+        // The other fills are actually registered by setting a background color
+        foreach ($this->registeredFills as $styleId) {
+
+            /** @var  $style Style */
```
adrilo: should be `/** @var Style $style */`

tests/Spout/Writer/XLSX/WriterWithStyleTest.php
```diff
@@ -237,6 +236,34 @@ public function testAddRowWithStyleShouldApplyWrapTextIfCellContainsNewLine()
...
+    public function testAddBackgroundColor()
+    {
+        $fileName = 'test_default_background_color.xlsx';
+        $dataRows = [
+            ["defaultBgColor"],
+        ];
+        $style = (new StyleBuilder())->setBackgroundColor(Color::WHITE)->build();
+        $this->writeToXLSXFile($dataRows, $fileName, $style);
+        $fillsDomElement = $this->getXmlSectionFromStylesXmlFile($fileName, 'fills');
+        $this->assertEquals(3, $fillsDomElement->getAttribute('count'), 'There should be 3 fills, including the default one.');
```
adrilo: including the 2 default ones

src/Spout/Writer/XLSX/Helper/StyleHelper.php
```diff
@@ -14,6 +15,64 @@
 class StyleHelper extends AbstractStyleHelper
 {
     /**
+     * @var array
+     */
+    protected $registeredFills = [];
```
adrilo: yes, totally belongs here! :)

adrilo:
👍 This looks great!

Now regarding your comments:
> Wrote better tests

Feel free to do it now or later. I'm comfortable pushing your changes as is.

> Deleted bgColor from the XLSX definition. Openpyxl does not use it when just a "start" color is supplied, so I decided that they have already solved this problem. When we decide to introduce an "end" color - we can introduce it again.

Sounds good!

> Hacked the two default styles by using a seperate counter for the fillIndex.

Yeah I guess that's fine. Another way to do it is to store the "pattern" and "fgcolor" attributes in an array inside `registeredFills`. This way you can add the 2 default fills to this array when it gets initialized and you treat all fills the same way when you render the fills content.

Thanks again for this pull request! I'll merge it whenever you tell me you're done with it.

madflow:
I have pushed the changes, squashed the commits and rebased.

adrilo:
Sorry, I just merged a new commit. I did not see you had pushed a new commit... Can you rebase one more time please?

BTW do you know if I can do it for you? It's annoying to have to ask you every time...

madflow:
I just rebased... I do not think that it is possible as a "Hub" feature. I can give your write access to my fork - but than you would be just  doing my work. But - not 100% sure. Would be interesting to know!

adrilo:
Thanks again @madflow !

PR #220, opened 2016-05-18
Retrieve XLSX sheets in order of appearance

### LINKED ISSUE #207: Sheets iterating by sheet name in alphabetical order
In the past when reading sheets they iterated in the order as shown in the excel file.  Now with the latest code I am finding that no matter what order the sheets are in the file the order in which they iterate is always alphabetical.  Is this a bug or by design?

adrilo:
Fixes #207 

Instead of relying on the ID, sheets should be retrieved in the order they appear in the file.
Workbook.xml describes the correct order.
This allows the reader to read data in the correct order when sheets have been manually moved after creation.


boxcla:

PR #221, opened 2016-05-19
Handle General number format as non date

### LINKED ISSUE #212: Numeric cells with automatic formatting being converted to date
[Numbers-Excel-Sample.xlsx](https://github.com/box/spout/files/258176/Numbers-Excel-Sample.xlsx)

Using the attached file (which was generated by Numbers on a Mac) you should be able to see 2 rows that look like the following:

Date | UniqueID | Latitude | Longitude | Value
---- | -------- | -------- | --------- | -----
2016-01-04 00:05:00 | 2776735 | 33.0195915 | -116.8402815 | 1

Using v2.4.4 of Spout, the second row is read incorrectly. The first cell is read correctly as a string. The second and subsequent cells are read as a numeric values, but are then converted to a date.

In the `isNumFmtIdCustomDateFormat` method, the `$numFmtId` specified is `0` which corresponds to `General`. This is then incorrectly detected as a date format with the regex check because the character `e` is in the word `General`.

I don't know much about the XLSX spec, but I think `General` should not be interpreted as a date. If I update the code to:

```php
$customNumberFormat = $customNumberFormats[$numFmtId];

//This is new code:
if ($customNumberFormat == 'General') {
    return false;
}
```

Then the cells are formatted as expected. Could you please take a look and see if you can implement a fix? Thanks!

adrilo:
Fixes #212 

If the number format is set to General (id = 0), do no try to format the value as a date

### DIFF
--- a/src/Spout/Reader/XLSX/Helper/StyleHelper.php
+++ b/src/Spout/Reader/XLSX/Helper/StyleHelper.php
@@ -171,9 +171,21 @@ class StyleHelper
     protected function doesNumFmtIdIndicateDate($numFmtId)
     {
         return (
-            $this->isNumFmtIdBuiltInDateFormat($numFmtId) ||
-            $this->isNumFmtIdCustomDateFormat($numFmtId)
+            !$this->doesNumFmtIdIndicateGeneralFormat($numFmtId) &&
+            (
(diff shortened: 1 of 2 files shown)


boxcla:

PR #224, opened 2016-05-19
Add support for cells formatted as time

### LINKED ISSUE #193: Can't read cell formatted as "time" in XLSX

adrilo:
Fixes #193 

Cells formatted as "time" have values between 0 and 1. These values used to be considered as invalid.
Note: this uses what was started in #202


boxcla:

PR #226, opened 2016-05-20
Option to return formatted dates instead of PHP objects

### LINKED ISSUE #178: Question about date type
When read xlsx cell with date value, result is DateTime object. It is fine if I expect date, but what if I expect some string and now I have to validate it. How to get original string, if I don't know original format?

For example if cell contains "31.10.1959" and it is date type, reader result is DateTime object with value "1959-10-31 00:00:00". Problem is that I can not show user that "1959-10-31" is not what I expect, because file contains "31.10.1959". Is there any way how to get original string without DateTime object? Or at least get original format?

adrilo:
Fixes #178 

When reading spreadsheets, Spout should be able to return formatted dates, as shown when opened with Excel for instance.
It currently only returns DateTime/DateInterval objects, making it impossible to read + write, as the Writer does not accept objects.


boxcla:

PR #227, opened 2016-05-20
Pin PHPUnit version to keep support for PHP5.4

adrilo:


boxcla:

PR #228, opened 2016-05-23
Update README.md

adrilo:
Fixed the amount of memory used by Spout, as it used to take into account the memory used by PHPUnit.


boxcla:

PR #229, opened 2016-05-23
XLSX Reader should add a space between text nodes

### LINKED ISSUE #219: XLSX Reader trims text nodes
The XLSX Reader trims cell values like this:

```
<si>
    <r>
        <rPr><sz val="10"/><rFont val="Arial"/><family val="2"/><charset val="1"/></rPr>
        <t>go to
        </t>
    </r>
    <r>
        <rPr><sz val="10"/><color rgb="FF0000FF"/><rFont val="Arial"/><family val="2"/><charset val="1"/></rPr>
        <t>https://github.com</t>
    </r>
    <r>
        <rPr><sz val="10"/><rFont val="Arial"/><family val="2"/><charset val="1"/></rPr>
        <t>
            please</t>
    </r>
</si>
```

but not like this:

```
<si>
    <t xml:space="preserve">text</t>
</si>
```

[sheet_with_hyperlinks.xlsx](https://github.com/box/spout/files/270022/sheet_with_hyperlinks.xlsx)

This results in:

```go tohttps://github.complease```

instead of:

```go to https://github.com please```

The ```preserve``` attribute is kind of vague: https://www.w3.org/TR/xml/#sec-white-space ```the value "preserve" indicates the intent that applications preserve all the white space.```

I do not believe, that in its absence the whitespaces, carriage returns, tabs etc. should be stripped from start and end of the string. Actually: all whitespace characters should be stripped if this is wanted behaviour.

```This specification does not give meaning to any value of xml:space other than "default" and "preserve". It is an error for other values to be specified; the XML processor MAY report the error or MAY recover by ignoring the attribute specification or by reporting the (erroneous) value to the application. Applications may ignore or reject erroneous values.```

Imho we should

* ignore it (preferred by me)
* introduce a config flag for this

adrilo:
Fixes #219 

When a cell contains multiple text nodes, the cell value is currently obtained by concatenating the value of each text node.
Instead, values should still be concatenated but a space should be added in between.


boxcla:

adrilo:
@madflow fyi

Issue #230, opened 2016-05-24
Unable to recover original row index from XLSX reader


KiNgMaR:
Hi there @adrilo et al,

in an XLSX file, completely empty rows seem to be omitted from the XML altogether. This means that for a sheet like the following example

    1 | Title  |      |
    2 |        |      |
    3 | Data   | Data |
    4 | Data   | Data |

the indices returned from `$reader->getSheetIterator()` will be assigned like this:

    0 | Title
    1 | Data
    2 | Data

which means that it's not possible to recover the original row positions (indices). This is an issue if you want to point to users to specific rows / cells in their sheets.

One possible solution would be to modify the `RowIterator` so that the index (key) returned matches the actual row number, i.e.:

    1 | Title
    3 | Data
    4 | Data

However, this clearly is a BC break, so before I make a PR, I would like to hear your opinion on this one... Thanks!

adrilo:
Hey @KiNgMaR,

As you noticed, Spout ignores empty lines. I understand that this may not be the best behavior.
A non-breaking change can be implemented, adding the option to choose whether empty lines should be kept or ignored. This would result in:

```
0 | Title
1 | 
2 | Data
3 | Data
```

It still not what you're proposing. I'm not sure if it's better to have the row index be the row number in the spreadsheet or just an indicator that has nothing to do with the spreadsheet's structure.

If anyone has a preference, please speak up :)

madflow:
I would prefer an option to preserve empty lines.

KiNgMaR:
Hey,

the root issue is not really that spout actively _skips_ empty rows, it's that empty rows do not even appear in the XML:

    <row r="1" x14ac:dyDescent="0.2" ht="15.75" spans="1:17">
        ...
    </row>
    <row r="3" x14ac:dyDescent="0.2" spans="1:17">
        ...

Maybe we can have two options? One that preserves empty lines (also relevant for the other readers) and one that uses the actual row indices as keys for the iterator (not relevant for CSV). Then again these may be too many options... hence my original proposal to simply use the row indices as keys :)

One alternative would be to add the row index somewhere inside the returned row. But as it's an array, using something like _rowindex feels weird and also breaks BC in a most horrible way.

Rushleader:
I like the idea of the option that preserves the row indexing (and adds empty rows where required). Should be the cleanest way. 
Default option should be just load the rows and skip empty I think

adrilo:
Thanks for your input! Very valuable :)
I believe the default should be:
[...]

PR #235, opened 2016-05-27
Preserve row indices [for discussion]

KiNgMaR:
This is a shot at #230, so we have something to discuss.

Summary:
* added new `ReaderOptions` class (for libraray-internal use) to avoid excessive arguments lists
* added new `shouldPreserveRowIndices` option + tests, implementation for XLSX, ODS

Missing:
* implementation of `shouldSkipEmptyRows` option. This is a bit tricky because of the buffering/streaming nature of the readers. Will work on it based on our discussion :smiley_cat: 

P.S. It's not fully cleaned up / ready for merge. Let's discuss if it's a sensible approach or not :)


adrilo:
I definitely like this approach. I've been tempted to create the Reader/WriterOptions for a while but I never did it. With a few more options available, I think it's good to have it. I share your opinion for keeping it internal as I like to have clean and discoverable APIs.

Based on my last comment on #230, I'd rather have the `shouldPreserveRowIndices` as the default behavior. When iterating over the rows, PHP will handle missing indices perfectly so it does not really matter if the array is [1,2,3] or [1,3,5], as long as the rows are in the correct order.

Also, the option `shouldOmitEmptyRows` should actually be `shouldPreserveEmptyRows` as the default behavior should be to skip empty rows. I agree that it's a bit tricky but should be doable :) Let's make sure we agree on the direction to take before starting the implementation.

KiNgMaR:
Pull request updated to reflect latest discussion, but I'm not satisfied with the XLSX RowIterator implementation yet...

src/Spout/Reader/ODS/RowIterator.php
```diff
@@ -130,23 +139,23 @@ public function next()
...
+                        // In Excel, the number of supported columns is 16384, but we don't want to returns rows with
+                        // always 16384 cells.
+                        if ((count($rowData) + $numColumnsRepeated) !== self::MAX_COLUMNS_EXCEL) {
+                            for ($i = 0; $i < $numColumnsRepeated; $i++) {
+                                $rowData[] = $cellValue;
+                            }
                         }
-                        $this->numReadRows++;
+                    } elseif ($this->readerOptions->shouldPreserveEmptyRows()) {
+                        $rowData = array_fill(0, $numColumnsRepeated, '');
```
madflow: This will create the "a lot of empty cells". For example if you take the test ```ODS\ReaderTest::testReadWithFilesGeneratedByExternalSoftwares``` and insert:

```    
public function testReadWithFilesGeneratedByExternalSoftwares($fileName, $skipLastEmptyValues)
{
        $reader = ReaderFactory::create(Type::ODS);
        /** @var  $reader Reader */
        $reader->setShouldPreserveEmptyRows(true);

        $allRows = $this->getAllRowsForFile($fileName, $reader);
 // ....
```

you will get 16384 empty cells in the "preserved" empty row.  I am not sure how to circumvent this - but this does not sound "right" ;)

madflow:
I know you asked for comments on the approach - so apologies if this is to premature: 

[consecutive_empty_rows_libreoffice.ods.zip](https://github.com/box/spout/files/294132/consecutive_empty_rows_libreoffice.ods.zip)

I tinkered with the ODS reader with the code from this PR and came across the artifact, that the reader does not preserve consecutive empty rows in a spreadsheet created by LibreOffice:

```

$reader = ReaderFactory::create(Type::ODS);
/* @var $reader \Box\Spout\Reader\XLSX\Reader */
$reader->open('consecutive_empty_rows_libreoffice.ods');
$reader->setShouldPreserveEmptyRows(true);

foreach ($reader->getSheetIterator() as $sheetIndex => $sheet) {
    foreach ($sheet->getRowIterator() as $row) {
        $rows[] = $row;
    }
}
$reader->close();

var_dump($rows);

array(3) {
  [0] =>
  array(1) {
    [0] =>
    string(5) "First"
  }
  [1] =>
  array(1) {
    [0] =>
    string(0) ""
  }
  [2] =>
  array(1) {
    [0] =>
    string(6) "Second"
  }
}
```

The xlsx reader does it correctly (saving the file as xlsx from LibreOffice):

```
array(4) {
  [0] =>
  array(1) {
    [0] =>
    string(5) "First"
  }
  [1] =>
  array(1) {
    [0] =>
    string(0) ""
  }
  [2] =>
  array(1) {
    [0] =>
    string(0) ""
  }
  [3] =>
  array(1) {
    [0] =>
    string(6) "Second"
  }
}
```

KiNgMaR:
Thanks for the support, I have been aware that consecutive empty rows needed testing, so this is very helpful :) Will continue amending the PR next week!

KiNgMaR:
Looks like the ODS reader needs support for `rows-repeated`. :)

KiNgMaR:
PR updated, looking forward to your continued feedback :smiley:

KiNgMaR:
Any feedback? Shall I begin to tidy up this version? Possibly split into multiple PRs?

src/Spout/Reader/ODS/RowIterator.php
```diff
@@ -161,7 +196,9 @@ public function next()
             throw new IOException("The sheet's data cannot be read. [{$exception->getMessage()}]");
         }
 
-        $this->rowDataBuffer = $rowData;
+        for ($i = 0; $i < $numRowsRepeated; ++$i) {
```
adrilo: This will be a problem. If you have one row repeated 100,000 times, you will have memory problems as you'll end up storing 100,000 rows in memory.

src/Spout/Reader/ODS/RowIterator.php
```diff
@@ -99,15 +108,34 @@ public function valid()
      */
     public function next()
     {
+        $prevRow = null;
+
+        if (count($this->rowDataBuffer) > 1) {
```
adrilo: Can you quickly explain the logic for this change? I'm having trouble understanding how the prevRow/rowDataBuffer work together

KiNgMaR: I agree it's too intricate, I'll investigate if it's possible to use a solution where only one row is buffered (which would also address your other comment regarding the repeated rows).

src/Spout/Reader/ODS/RowIterator.php
jmleroux: :+1: for @adrilo 
Plus, the `next()` call will return more than one row as if the iterator had read multiples rows.

adrilo:
Fixed in #336

KiNgMaR:
Thanks you so much for picking this up! I tested on master and it works great for my use case as well 💯

PR #237, opened 2016-05-30
Support XLSX with prefixed XML files

### LINKED ISSUE #233: Cannot read XLSX file
Hello!
I just want to read XLSX file (in attachment), but got the error:
Notice: Undefined offset: 0 in W:\home\local\test\library\Spout\Reader\XLSX\Helper\SheetHelper.php on line 109 Fatal error: Call to a member function getAttribute() on a non-object in W:\home\local\test\library\Spout\Reader\XLSX\Helper\SheetHelper.php on line 111

Spout version - 2.4.4

[1.xlsx](https://github.com/box/spout/files/285193/1.xlsx)

adrilo:
Fixes #233 

While the standard is not to have prefixes, some XLSX files have XML files containing a prefix.
Microsoft has a tool that generates such files: https://msdn.microsoft.com/en-us/library/office/gg278316.aspx


boxcla:

PR #239, opened 2016-05-30
Convert escapers to singletons

### LINKED ISSUE #236: Helper classes should be singletons
StyleHelper, CellValueFormatter, etc... don't need to be created more than once, as they only expose helper "static" functions. They should be converted to singletons to improve memory usage.

adrilo:
Fixes #236


boxcla:

PR #255, opened 2016-06-16
Fix XLSX reading when shared strings missing the uniqueCount attribute

### LINKED ISSUE #254: Can't support xlsx file created by WPS Office(an office software likes MS office, almost only used in China)
For some reason, my program should read the data from xlsx file which creates by WPS Office.

I was disturbed by the difference between WPS Office and MS office.

At first, all my test xlsx file was from WPS. I got the error as follow.

> <b>Fatal error</b>:  Uncaught exception 'Box\Spout\Common\Exception\IOException' with message 'Could not open C:/Users/sf/Desktop/Book1.xlsx for reading! (Index invalid or out of range)' in D:\AppServ\www\exp\spout\src\Spout\Reader\AbstractReader.php:105
> Stack trace:
> #0 D:\AppServ\www\exp\test_spout.php(22): Box\Spout\Reader\AbstractReader-&gt;open('C:/Users/sf/Des...')
> #1 {main}
>   thrown in <b>D:\AppServ\www\exp\spout\src\Spout\Reader\AbstractReader.php</b> on line <b>105</b><br />

But after I transfer Book1.xlsx from WPS one to MS one, I get the result I want!

I really want to use the library, to low the memory usage(That attract me!).

Moreover, WPS Office has enough user in China, just like MS Office in other country, I think.

So, if you could improve the library to support xlsx file from WPS, it will make Spout be more popular in Chinese developer.
And personally, I will be very thankful if you could do that.

adrilo:
Fixes #254 

Use "count" attribute as a fallback


boxcla:

PR #268, opened 2016-07-11
Add support for missing cell reference

### LINKED ISSUE #267: XLSX: error parsing file with missing "r" cell attributes
I have a problem reading an automatically-generated XLSX file. The file was sent from a Microsoft-owned product, so I assume its format does not violate XLSX specification... although I may be wrong :disappointed: 

I can open the file manually in Libre Office, but Spout throws an exception with a "Cannot get column index from an invalid cell index" message. After a bit of digging I found that error is caused by sheet's XML file (`sheet1.xml`) - row and cell elements do not contain "r" attributes, so `CellHelper::getColumnIndexFromCellIndex` method is called with NULL parameter.

The bug is quite easy to reproduce:
- create a simple XLSX spreadsheet (even filling a single cell is enough)
- unpack the XLSX file
- open the `xl/worksheets/sheet1.xml` file in unpacked archive
- delete "r" attribute from at least one `<c>` element
- save changes and repack the XLSX file
- try to read the modified file in Spout

I attach such simple file as an example: [test.xlsx](https://github.com/box/spout/files/356972/test.xlsx)

adrilo:
Fixes #267

When describing a cell, the cell reference (r="A1") is optional.
When not present, we should just increment the index of the last processed row.


boxcla:

aistis-:
Cool, could you tag a new version with this fix? :)

adrilo:
2.5.0 should contain this fix already

PR #269, opened 2016-07-11
Shared strings table without uniqueCount and count should work

### LINKED ISSUE #260: Problems reading XLSX files generated by Crystal Reports
I'm using the latest master version and I'm getting this exception with an XLSX file exported by Crystal Reports:
Box\Spout\Common\Exception\IOException with message:
Could not open test.xlsx for reading! (Index invalid or out of range)
in Spout/Reader/AbstractReader.php line 105

It could be related to:
https://github.com/box/spout/issues/254

If I open the xlsx file in LibreOffice and save it again, then it works fine. 

When reading other Crystal Reports xlsx files, I had other weird things happen such as custom format strings being displayed instead of the actual values in some fields, and in some other fields empty strings were returned when there were really values in those fields. But this may show up if the reported exception problem is solved.
Is there a way to send the small file privately because it contains real production (possibly somewhat sensitive) data? I get the files by proxy, so that's why I can't easily get test data.

adrilo:
Fixes #260 

Use file based strategy in this case


boxcla:

PR #275, opened 2016-07-14
Do not return anything from constructor

adrilo:


boxcla:

PR #282, opened 2016-07-21
XLSX cells containing date values should respect shouldFormatDate option

### LINKED ISSUE #279: CellValueFormatter::formatDateCellValue() doesn't respect $shouldFormatDates
There is not check in `\Box\Spout\Reader\XLSX\Helper\CellValueFormatter::formatDateCellValue` if `$shouldFormatDates` is enabled. 

I don't want to convert dates to PHP `DateTime`. How can I disable it?

adrilo:
Fixes #279 

Return the ISO 8601 date string directly if option is set


boxcla:

PR #287, opened 2016-08-10
Spout should be able to read prefixed styles.xml

### LINKED ISSUE #283: Read date from xlsx
I have this file: [nm_sd_apturesana_21072016.xlsx](https://github.com/box/spout/files/385702/nm_sd_apturesana_21072016.xlsx)
I can not understand why spout returns:
```
array (48) [
    string (2) "1."
    string (11) "40003691882"
    string UTF-8 (45) "Sabiedrība ar ierobežotu atbildību "PRINTECH""
    integer 42570
    integer 42572
    ...
]
```
Row contains two dates "2016.07.19" and "21.07.2016". Both cells are with date format so why it returns integers "42570" and "42572". And how I can get correct date from them?

adrilo:
Fixes #283 

`styles.xml` can also be prefixed. SimpleXML was used to parse style data but it does not handle prefixes by default. Therefore changing to Spout's XMLReader parsing that has prefix support built-in

### DIFF
--- a/src/Spout/Reader/XLSX/Helper/StyleHelper.php
+++ b/src/Spout/Reader/XLSX/Helper/StyleHelper.php
@@ -2,7 +2,6 @@
 
 namespace Box\Spout\Reader\XLSX\Helper;
 
-use Box\Spout\Reader\Wrapper\SimpleXMLElement;
 use Box\Spout\Reader\Wrapper\XMLReader;
 
 /**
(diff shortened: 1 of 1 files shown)


boxcla:

DMGPage:
When we will see this bug fix in stable version release?

adrilo:
@DMGPage Now :) I just released v2.6.0

PR #290, opened 2016-08-17
API to set default row style

### LINKED ISSUE #272: Setting default row style
Is there a way to override the default row style? I've looked through the code and it appears to be all done via constants and protected functions so I'm guessing not. For example, if I want to use Arial 11 for all cells in an XLSX file, and the existing default is Calibri 12, then using AddRowWithStyle for every row for tens/hundreds of thousands of rows is somewhat inefficient due to all of the setRowStyle / resetRowStyleToDefault calls which are wrapped around each addRow call.

I know I could use AddRowsWithStyle however this would mean having to create an array with all of the row data first, and in the case of hundreds of thousands of rows this could have an impact on PHP memory usage which is what I'm trying to avoid - for example if the original data is coming from a database then reading it out row by row and writing to XLSX will consume a lot less memory than reading all of the data into an array first.

adrilo:
Fixes #272

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,6 +185,19 @@ $writer->addRowWithStyle(['Border Bottom Green Thin Dashed'], $style);
 $writer->close();
 ```
 
+Spout will use a default style for all created rows. This style can be overridden this way:
+
+```php
+$defaultStyle = (new StyleBuilder())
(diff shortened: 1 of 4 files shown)


boxcla:

PR #299, opened 2016-08-30
Fix #297

### LINKED ISSUE #297: Fill/Background definitions are not properly reused
When two styles share the same background definition - the existing fillId is not properly reused. Like here in this comment in the "borders" PR:

https://github.com/box/spout/pull/289#discussion_r74831820

madflow:
Fixes #297


boxcla:

adrilo:
👍

PR #304, opened 2016-09-01
Improved XLSX Escaper performance for strings with only common/safe characters.

nimmneun:
Improved XLSX Escaper performance for strings with only common/safe characters by about 10-20%, depending on the string/cellValue.

### DIFF
--- a/src/Spout/Common/Escaper/XLSX.php
+++ b/src/Spout/Common/Escaper/XLSX.php
@@ -95,6 +95,10 @@ class XLSX implements EscaperInterface
      */
     protected function escapeControlCharacters($string)
     {
+        if (1 !== preg_match('~[^A-Za-z0-9\-.:, ]~', $string)) {
+            return $string;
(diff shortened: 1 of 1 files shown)


boxcla:
Verified that @nimmneun has signed the CLA. Thanks for the pull request!

src/Spout/Common/Escaper/XLSX.php
```diff
@@ -95,6 +95,10 @@ protected function getControlCharactersEscapingMap()
      */
     protected function escapeControlCharacters($string)
     {
+        if (1 !== preg_match('~[^A-Za-z0-9\-.:, ]~', $string)) {
```
adrilo: This does not work for characters that are not in the a-Z range: accented characters, different alphabets, etc..

nimmneun:
closed b/c @adrilo solved this perfectly #305 :+1:

PR #307, opened 2016-09-07
Apply custom style to empty cells if needed

### LINKED ISSUE #295: Row style
Hi, It would be possible to apply the style to a field even if it is empty?

I have this:
```
$style = (new StyleBuilder())
		->setBackgroundColor("f2dede")
		->build();
	
		$writer->addRowWithStyle(['AAAAA','BBBB','','DDDDD','888888',null],$style);
```

But where the field is empty or is null, the style is not applied.

adrilo:
Fixes #295

If a row should be written with a custom style, the handling of empty cells should change.
Instead of being skipped entirely, empty cells will be applied the custom style, if this style has custom background color or borders.
If not, then the cell definition can still be skipped.

### DIFF
--- a/src/Spout/Writer/Style/Style.php
+++ b/src/Spout/Writer/Style/Style.php
@@ -288,7 +288,7 @@ class Style
 
     /**
      * Sets the background color
-     * @param $color ARGB color (@see Color)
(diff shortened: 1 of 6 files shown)


boxcla:

PR #311, opened 2016-09-09
Add a way to disable text wrapping through the api (Fix #247)

### LINKED ISSUE #247: there is no way to disable textWraping through the API
since enhancement https://github.com/box/spout/pull/83
Spout automatically wraps cells that contains newlines and we want to disable that, but current API does not provide any means for that.

Style builder only has an option to enable wraping but not to disable.
`Style->shouldWrapText` if only set to true by setShouldWrapText method. 

Perhaps you could modify that method and add a parameter that would allow to toggle this option?

Hastegan:
Fixes #247

### DIFF
--- a/src/Spout/Writer/Style/Style.php
+++ b/src/Spout/Writer/Style/Style.php
@@ -269,7 +269,7 @@ class Style
     }
 
     /**
-     * @param boolean $shouldWrap Should the text be wrapped
+     * @param bool $shouldWrap Should the text be wrapped
      * @return Style
      */
     public function setShouldWrapText($shouldWrap = true)
@@ -280,7 +280,7 @@ class Style
     }
 
     /**
-     * @return boolean
+     * @return bool
      */
     public function hasSetWrapText()
     {
--- a/src/Spout/Writer/Style/Style.php
+++ b/src/Spout/Writer/Style/Style.php
@@ -269,7 +269,7 @@ class Style
     }
 
     /**
-     * @param bool $shouldWrap Should the text be wrapped
+     * @param bool|void $shouldWrap Should the text be wrapped
      * @return Style
      */
     public function setShouldWrapText($shouldWrap = true)
--- a/src/Spout/Writer/Style/Style.php
+++ b/src/Spout/Writer/Style/Style.php
@@ -269,13 +269,22 @@ class Style
     }
 
     /**
+     * @param boolean $shouldWrap Should the text be wrapped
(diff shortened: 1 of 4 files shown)


boxcla:
Hi @Hastegan, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

Hastegan:
CLA signed

boxcla:
Verified that @Hastegan has just signed the CLA. Thanks, and we look forward to your contribution.

src/Spout/Writer/Style/StyleBuilder.php
```diff
@@ -109,15 +109,15 @@ public function setFontName($fontName)
     }
 
     /**
-     * Makes the text wrap in the cell if it's too long or
-     * on multiple lines.
+     * Makes the text wrap in the cell if requested
      *
      * @api
+     * @param boolean $shouldWrap Should the text be wrapped
```
adrilo: `@param bool` (for consistency, as we've been using "bool" instead of "boolean" everywhere and this is what PHP uses too)

src/Spout/Writer/Style/Style.php
```diff
@@ -269,16 +269,25 @@ public function shouldWrapText()
...
+    public function setShouldWrapText($shouldWrap = true)
     {
-        $this->shouldWrapText = true;
+        $this->shouldWrapText = $shouldWrap;
         $this->hasSetWrapText = true;
         return $this;
     }
 
     /**
+     * @return boolean
```
adrilo: `@param bool` (for consistency, as we've been using "bool" instead of "boolean" everywhere and this is what PHP uses too)

src/Spout/Writer/Style/Style.php
```diff
@@ -269,16 +269,25 @@ public function shouldWrapText()
     }
 
     /**
+     * @param boolean $shouldWrap Should the text be wrapped
```
adrilo: `@param bool|void` (for consistency, as we've been using "bool" instead of "boolean" everywhere and this is what PHP uses too)

Hastegan: I have noticed that almost all PHPDoc in this file is written with 'boolean' instead of 'bool'

adrilo: Indeed... It should be "bool"

PR #314, opened 2016-09-16
Fix boolean notation in PHPDoc

Hastegan:
As pointed out in #311 , booleans in PHPDoc should be written as 'bool', not 'boolean'.


boxcla:
Verified that @Hastegan has signed the CLA. Thanks for the pull request!

adrilo:
Thanks for following-up with this commit!

PR #316, opened 2016-09-24
Better support for Date custom format

### LINKED ISSUE #312: Unable to read dates with some XLSX files
The `cellXfs` section defined in `styles.xml` describe some of the transformations that need to be applied to the cells. To determine whether a cell contains a date, it currently needs to have 2 things:
- an attribute called `applyNumberFormat` set to "true"
- an attribute called `numFmtId` reference the date format

However, sometimes the `applyNumberFormat` is missing (but `numFmtId` is there). This causes the date formatting to be ignored...

An example can be found here: [tests.xlsx](https://github.com/box/spout/files/468072/tests.xlsx)

adrilo:
Fixes #312 

- To determine if a style should apply a date format, the presence of "applyNumberFormat" attribute on the "cellXfs" section of styles.xml is now optional. We only look at the "numFmtId" attribute (but early return if "applyNumberFormat" is set to "0").
- The format code can contain lowercase AND now uppercase characters as its pattern.
- "General" format code used as a custom format is now supported. It seems to be used by a bunch of programs...


boxcla:

PR #320, opened 2016-09-30
Removing license badge

adrilo:
As Github now provides this information in the project's header


boxcla:

PR #327, opened 2016-10-11
Temp files should be deleted when an exception is thrown

### LINKED ISSUE #315: Please delete the temporary file
Spout does not seem to delete the temporary files used to generate `.xlsx` exports.

This has caused a bit of trouble for us, as we ran out of space.

adrilo:
Fixes #315

If an exception is thrown while writing data, instead of letting the developer handle this situation gracefully, Spout can attempt to delete all the temporary files that were created so far, as well as the output file as it won't be completed and therefore corrupted.


boxcla:

PR #328, opened 2016-10-12
Fix output file deletion after exception thrown on write

adrilo:
For relative paths, it would not work as the FileSystemHelper would not allow deleting a file that's not part of its base folder.


boxcla:

PR #331, opened 2016-10-13
Option to keep empty rows

### LINKED ISSUE #230: Unable to recover original row index from XLSX reader

adrilo:
Fixes #230

This change adds a new setting to readers: preserving empty rows. Currently, Spout skips empty rows but for some users, it may be useful to keep them. This is now possible with all readers.


boxcla:

adrilo:
FYI @KiNgMaR @madflow @Rushleader

madflow:
I did a quick test and this will work for my use case for this feature. Thanks! 

Maybe worth noticing that the options setters have to be called before the reader is opened.

Also: The ods reader/writer created a different result with my test code:

https://gist.github.com/madflow/c2ccef2409fac04544cf9c7537bbcfb6

adrilo:
Cool, thanks for taking a look @madflow !

Issue #329, opened 2016-10-12
Vertical tabs break ODS files


madflow:
\v breaks the ODS writer

```php
$writer = \Box\Spout\Writer\WriterFactory::create(Type::ODS);
$writer->openToFile(__DIR__ . '/failed.ods');
$writer->addRow(["\v"]);
$writer->close();
```

> Read-Error.
Format error discovered in the file in sub-document content.xml at 9,387(row,col)

Tested with LibreOffice 5.2.

CSV and XLSX work fine.

Should we fix this in upstream (spout)?

This came up from user input, probably copy-pasted from somewhere.

adrilo:
Yes, it'd be nice to have a fix in Spout, so that other developers would benefit from it.
Feel free to submit a pull request!

BTW, is it broken with other characters too?

madflow:
A vertical tab is a illegal character in this xml spec. 

http://stackoverflow.com/questions/1091945/what-characters-do-i-need-to-escape-in-xml-documents
https://www.w3.org/TR/xml/#charsets

I also found, that a form feed breaks the XML.

There already seems to be an escaping mechanism for XLSX content with some heavy preg_replace'ing - maybe we can adapt something form there.


```php
require __DIR__ . '/vendor/autoload.php';

use Box\Spout\Reader\ReaderFactory;
use Box\Spout\Common\Type;
use Box\Spout\Writer\WriterFactory;

$writerTypes = [Type::ODS, Type::XLSX];

$chars = ["\v", "\f", "<", ">", "&", ","];

foreach($writerTypes as $writerType) {

    foreach($chars as $index => $char) {

        $writer = WriterFactory::create($writerType);
        $fileName = sprintf('/invalid_chars_%s.%s', $writerType, $writerType);
        $writer->openToFile(__DIR__ . $fileName);
        $writer->addRow([$char]);
        $writer->close();

        $reader = ReaderFactory::create($writerType);
        $reader->open(__DIR__ . $fileName);

        try {
            /** @var \Box\Spout\Reader\XLSX\Sheet $sheet */
            foreach($reader->getSheetIterator() as $sheet) {
                foreach($sheet->getRowIterator() as $row) {

                }
            }
        } catch(\Exception $e) {
            echo sprintf('The char at index %s is invalid in writer %s', $index, $writerType).PHP_EOL;
        }

        $reader->close();

    }
}
```

> The char at index 0 is invalid in writer ods
> The char at index 1 is invalid in writer ods

[...]

PR #333, opened 2016-10-17
Fix issue 329

madflow:
Tries to do some good for #329

### DIFF
--- a/src/Spout/Common/Escaper/ODS.php
+++ b/src/Spout/Common/Escaper/ODS.php
@@ -22,7 +22,7 @@ class ODS implements EscaperInterface
      */
     public function escape($string)
     {
-        return htmlspecialchars($string, ENT_QUOTES);
+        return htmlspecialchars($string, ENT_QUOTES | ENT_DISALLOWED);
     }
 
     /**
--- a/src/Spout/Common/Escaper/ODS.php
+++ b/src/Spout/Common/Escaper/ODS.php
@@ -4,9 +4,6 @@ namespace Box\Spout\Common\Escaper;
 
 use Box\Spout\Common\Singleton;
 
-// HVM hack https://github.com/box/spout/issues/329
-defined('ENT_DISALLOWED') || define('ENT_DISALLOWED', 128);
-
 /**
  * Class ODS
  * Provides functions to escape and unescape data for ODS files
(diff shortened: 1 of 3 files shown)


boxcla:

tests/Spout/Writer/ODS/WriterTest.php
```diff
@@ -4,6 +4,7 @@
 
 use Box\Spout\Common\Exception\SpoutException;
 use Box\Spout\Common\Type;
+use Box\Spout\Reader\ReaderFactory;
```
adrilo: Let's avoid using the Reader in the Writer test. Instead, can you add a test in the ODSTest.php file that tests the escaper? Although it won't test the same thing (the test you wrote being a full integration testing, the escaper test being more like a unit test), I think it will be good enough.

src/Spout/Common/Escaper/ODS.php
```diff
@@ -22,7 +22,15 @@ class ODS implements EscaperInterface
...
     public function escape($string)
     {
-        return htmlspecialchars($string, ENT_QUOTES);
+        if (defined('ENT_DISALLOWED')) {
+            return htmlspecialchars($string, ENT_QUOTES | ENT_DISALLOWED);
+        } else {
+            // We are on hhvm or any other engine that does not support ENT_DISALLOWED
+            // https://github.com/box/spout/issues/329
+            $escapedString =  htmlspecialchars($string, ENT_QUOTES);
+            $replacedString = preg_replace('/[\x00-\x08\x0B-\x0C\x0E-\x1F]/', '�', $escapedString);
```
adrilo: can you define a extra variable for the regex pattern. Similar to what's done for the XLSX escaper:

```php
// control characters values are from 0 to 1F (hex values) in the ASCII table
// some characters should not be escaped though: "\t", "\r" and "\n".
return '[\x00-\x08' .
        // skipping "\t" (0x9) and "\n" (0xA)
        '\x0B-\x0C' .
        // skipping "\r" (0xD)
        '\x0E-\x1F]';
```

adrilo:
Few minor comments, but looks great overall :)

madflow:
I shamelessly copy-pasted from the XLSX escaper and test... This should be a subject for future refactoring I suppose.

adrilo:
:) That's fine for now! Thanks for the fix.

PR #335, opened 2016-10-17
ODS Reader should support num-rows-repeated for non empty rows

adrilo:


boxcla:

PR #338, opened 2016-10-18
Add ReaderOptions for all readers

adrilo:
Instead of passing every single option down the chain, pass an object around


boxcla:

PR #341, opened 2016-10-18
Remove text suffix in XLSX date formats

### LINKED ISSUE #330: Option to show datetimes with the original format
Hi!
I've some problems when I process a xlsx with spanish dates.

I work with spanish dates like 'D-M-Y' (28-12-2017) or 'D/M/Y' (28/12/2017) and the setShouldFormatDates option is a headache for me.

There are any possibility to add an option to return the string that is showed in the xlsx when the cell is a date instead of return an object or the date in english format (or xlsx date format) ??

I need to extract the date with the original format, not with a different format or an object.

Thanks for this great package!

adrilo:
Fixes #330

Some date formats have a text suffix, e.g. "mm/dd/yy;@". We should remove the ";...@" part.


boxcla:

PR #354, opened 2016-11-03
Fix crash when using associative array with empty row

### LINKED ISSUE #353: `Worksheet::addRow($row)` does not accept associative arrays anymore.
`Worksheet::addRow($row)` now calls `Worksheet::isEmptyRow($dataRow)`) that expects `$row` to be an array with numeric, sequential, zero-based keys.

Issue introduced in v2.7.0 (changing behaviour from v2.6.0, that used to accept associative arrays).

adrilo:
Fixes #353


boxcla:

PR #364, opened 2016-11-29
Improve error message when invalid sheet name set

### LINKED ISSUE #362: Change request to make InvalidSheetNameException more descriptive
**Box\Spout\Writer\Common\Sheet.php**

**Background:**
I was setting sheet names to year month periods such as 201611 (an int) and getting InvalidSheetNameException exceptions. I traced it down to being the !== operator in doesContainInvalidCharacters() which may be something for a different issue report.

**Change request:**
I could create a fork and a pull request, but this is fairly short...

Replace first 2 lines of setName() with:
```php
if (!$this->isNameValid($name, $error_reason)) {
            $errorMessage = "The sheet's name is invalid ($error_reason). It did not meet at least one of these requirements:\n";
```
Note the $error_reason which receives the reason and prevents unnecessary hair pulling.

Replace isNameValid() with:
```php
    /**
     * Returns whether the given sheet's name is valid.
     * @see Sheet::setName for validity rules.
     *
     * @param string $name
     * @param string &$error_reason optional, receives reason on error.
     * @return bool TRUE if the name is valid, FALSE otherwise.
     */
    protected function isNameValid($name, &$error_reason = null)
    {
        if (!(is_string($name) || is_int($name))) {
        	$error_reason = 'not a string or int';
            return false;
        }
        $nameLength = $this->stringHelper->getStringLength($name);
        if (!$nameLength) {
        	$error_reason = 'zero length';
        	return false;
        }
        if ($nameLength > self::MAX_LENGTH_SHEET_NAME) {
        	$error_reason = "length $nameLength is too long";
        	return false;
        }
        if ($this->doesContainInvalidCharacters($name)) {
        	$error_reason = 'contains invalid characters';
        	return false;
        }
        if (!$this->isNameUnique($name)) {
        	$error_reason = 'not unique';
        	return false;
        }
        if ($this->doesStartOrEndWithSingleQuote($name)) {
        	$error_reason = 'starts or ends with single quote';
        	return false;
        }
        return true;
    }
```

adrilo:
Fixes #362

Instead of listing all the requirements, only list the requirements that actually failed to be met.

### DIFF
--- a/src/Spout/Writer/Common/Sheet.php
+++ b/src/Spout/Writer/Common/Sheet.php
@@ -75,14 +75,7 @@ class Sheet
      */
     public function setName($name)
     {
-        if (!$this->isNameValid($name)) {
-            $errorMessage = "The sheet's name is invalid. It did not meet at least one of these requirements:\n";
-            $errorMessage .= " - It should not be blank\n";
-            $errorMessage .= " - It should not exceed 31 characters\n";
-            $errorMessage .= " - It should not contain these characters: \\ / ? * : [ or ]\n";
-            $errorMessage .= " - It should be unique";
-            throw new InvalidSheetNameException($errorMessage);
-        }
+        $this->throwIfNameIsInvalid($name);
 
         $this->name = $name;
         self::$SHEETS_NAME_USED[$this->index] = $name;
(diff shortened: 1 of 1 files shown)


boxcla:

PR #366, opened 2016-12-15
Make max read bytes per line configurable

a2xchip:
Hello! This PR introduces configurable property ```maxReadBytesPerLine``` and removes previously used constant.

- Remove max read bytes constant
- Update reader, reader options, row iterator adding max read bytes properties
- Add test data
- Update tests


boxcla:
Verified that @a2xchip has signed the CLA. Thanks for the pull request!

adrilo:
Hi @a2xchip! Thanks for the pull request.

I'm guessing this PR is solving a problem with CSV files containing very long lines (more than 32768 bytes). Is that correct?
We need to be really careful when introducing a new API, as it exposes more complexity to the end user. Maybe the default value can be increased to accommodate more use cases instead? If this solves 99.9% of the problem, I think that's a good tradeoff as it removes some complexity. What do you think?

a2xchip:
Hello @adrilo 
Yes, problem is that Spout is used as third-party dependency in Akeneo PIM. When PIM integration has a lot of languages and entity contains multiple translatable text fields, csv import fails.

IMHO there is no problem at all, cause we do not change current API but make it wider - no BC breaks and end users already using package will not feel it after upgrade.

SamirBoulil:
Hello,

I believe it is not in the library's best interest to extend the MAX_READ_BYTES_PER_LINE constant value because increasing this constant value will also increase the memory usage of every users of this library.

Also, we have very specific needs in our project that pushes us to read really large files. This new API will allow us to read those files while not changing the library's inner working for every one too munch.

a2xchip:
@boxcla Any news?

adrilo:
The constant was used when Spout was using `stream_get_line` (as it required a `length` value). The implementation was changed recently in favor of `fgetcsv`. `fgetcsv` does not require a length (or setting length to 0 accommodates for variable lengths).
So instead of adding an API, what about setting the length to 0? It will make the reading slower for very long lines but that's not the majority of the use cases. And the tradeoff for keeping the API simple seems reasonable.

What do you think? @a2xchip @SamirBoulil

adrilo:
@a2xchip @SamirBoulil ping!

adrilo:
@a2xchip Are you ok with this proposal? If so, can you update/create a new PR?

a2xchip:
@adrilo Hello! Thx for reply! Let me check it again first.

adrilo:
I created the PR: #399. Very simple change :)

a2xchip:
@adrilo Yeah, of course it is not hard to make change, but not sure if this changes is the best choice

adrilo:
I ran some perf tests with the PR I created and I did not see any perf regression. I used a CSV file with 500k lines, 140MB and the time taken to read the entire file is exactly the same.
Do you foresee any problems with this change?

a2xchip:
Sounds good!!! Let me close this PR and merge your once! @adrilo

PR #376, opened 2017-01-25
ensure libxml entity loading is enabled

kbond:
I had an issue where another part of my app was setting `libxml_disable_entity_loader(true)`. This caused failures when using spout. Would this change be acceptable?

### DIFF
--- a/src/Spout/Reader/Wrapper/XMLReader.php
+++ b/src/Spout/Reader/Wrapper/XMLReader.php
@@ -16,6 +16,12 @@ class XMLReader extends \XMLReader
 
     const ZIP_WRAPPER = 'zip://';
 
+    public function __construct()
+    {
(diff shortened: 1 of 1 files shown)


boxcla:
Hi @kbond, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

adrilo:
Hi @kbond,

Unfortunately, I can't merge that one as it changes a global state. If you application needs this to be disabled, Spout should not mess with this.
This needs to be handled case by case. Or maybe you can catch an error and return a meaningful error message, explaining that `libxml_disable_entity_loader` may need to be set to true?

kbond:
Ok, fair enough.

PR #381, opened 2017-02-06
Fix sheet name uniqueness bug when creating multiple workbooks.

bobwienholt:
Attempting to create two workbooks with similar sheet names in a different order causes an erroneus InvalidSheetNameException.

This commit uses the workbook sheet list to perform this check instead of a static variable in the Sheet class.

### DIFF
--- a/src/Spout/Writer/Common/Sheet.php
+++ b/src/Spout/Writer/Common/Sheet.php
@@ -164,7 +164,8 @@ class Sheet
      */
     protected function isNameUnique($name)
     {
-		foreach ($this->workbook->getWorksheets() as $sheetIndex => $sheet) {
+		foreach ($this->workbook->getWorksheets() as $sheetIndex => $worksheet) {
+			$sheet = $worksheet->getExternalSheet();
 			if ($sheetIndex !== $this->index && $sheet->getName() === $name) {
 				return false;
 			}
--- a/src/Spout/Writer/Common/Sheet.php
+++ b/src/Spout/Writer/Common/Sheet.php
@@ -3,6 +3,7 @@
 namespace Box\Spout\Writer\Common;
 
 use Box\Spout\Common\Helper\StringHelper;
+use Box\Spout\Writer\Common\Internal\WorkbookInterface;
 use Box\Spout\Writer\Exception\InvalidSheetNameException;
 
 /**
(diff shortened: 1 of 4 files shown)


boxcla:
Hi @bobwienholt, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

bobwienholt:
CLA signed

boxcla:
Verified that @bobwienholt has just signed the CLA. Thanks, and we look forward to your contribution.

seanbright:
This is biting me, can we get this merged?

adrilo:
@bobwienholt I'm away for 2 weeks but I'll take care of it as soon as I can. Thanks for your patience!

adrilo:
Looking at your code, I don't think it makes sense to do it the way you're doing it. A spreadsheet should have only one workbook (which can contain multiple sheets). I'm guessing that you're trying to create multiple spreadsheets in the same process and this is when the bug appears.

Someone else ran into the same issue (https://github.com/box/spout/issues/391) and the resolution seems more straightforward. @bobwienholt would that fix your problem or did I misunderstand something?

bobwienholt:
The fix in issue #391 seems to be "easier"... however, if you had a
situation where you had two different workbooks open at the same time, that
fix wouldn't be correct.  I had originally thought about taking that
approach to fix the problem but decided to go the extra mile and fix for
all possible scenarios.

It would seem to me that tracking the uniqueness of the sheets within the
workbook shouldn't be tracked using a static class variable.

I'd be happy to change my patch in anyway you feel would make it better.
Just let me know your thoughts.

- Bob

adrilo:
The solution proposed in the previously mentioned issue is actually incorrect, as it resets the value of the static variable every time a new sheet is created, erasing the names of all sheets already created...

I understand your point and agree that the uniqueness should be enforced at the workbook level. Your solution introduces a circular reference (workbook <=> sheet) and I'm not a big fan of that. Finding another way to achieve the same thing without the circular reference would be better :)

adrilo:
@bobwienholt Here is my take on it: #397
It's inspired by what you did, but instead of passing the reference to the workbook, I just pass an ID identifying this workbook. This way, `Sheet` does not need to call any functions on the workbook itself and can function independently.

adrilo:
Closing this as #397 has been merged

bobwienholt:
Awesome.  I really appreciate it.  Love this library by the way.

adrilo:
<3

PR #383, opened 2017-02-12
Cell value objects

madflow:
* #182 
* Introduces a "Cell" object for writers
* Streamlined the cell detection in ODS and XLSX writer to use the newly introduced Cell object
* Cell types are borrowed from Apache POI (for example https://poi.apache.org/apidocs/org/apache/poi/ss/usermodel/Cell.html)
* Some tests

🏷  RFC

### DIFF
--- a/src/Spout/Writer/Common/Cell.php
+++ b/src/Spout/Writer/Common/Cell.php
@@ -4,6 +4,11 @@ namespace Box\Spout\Writer\Common;
 
 use Box\Spout\Writer\Common\Helper\CellHelper;
 
+/**
+ * Class Cell
+ *
+ * @package Box\Spout\Writer\Common
+ */
 class Cell
 {
     /**
@@ -18,13 +23,14 @@ class Cell
 
     /**
      * Formula cell type
+     * Not used at the moment
      */
     const TYPE_FORMULA = 2;
 
     /**
-     * Blank cell type
+     * Empty cell type
      */
-    const TYPE_BLANK = 3;
+    const TYPE_EMPTY = 3;
 
     /**
      * Boolean cell type
@@ -38,20 +44,19 @@ class Cell
 
     /**
      * The value of this cell
-     * @var null | mixed
+     * @var mixed|null
      */
     protected $value = null;
 
     /**
      * The cell type
-     * @var null
+     * @var int
      */
     protected $type = null;
 
     /**
      * Cell constructor.
      * @param $value mixed
-     * @param $comment string
      */
     public function __construct($value)
     {
@@ -92,7 +97,7 @@ class Cell
         if (CellHelper::isBoolean($value)) {
             return self::TYPE_BOOLEAN;
         } elseif (CellHelper::isEmpty($value)) {
-            return self::TYPE_BLANK;
+            return self::TYPE_EMPTY;
         } elseif (CellHelper::isNumeric($this->getValue())) {
             return self::TYPE_NUMERIC;
         } elseif (CellHelper::isNonEmptyString($value)) {
@@ -113,12 +118,13 @@ class Cell
     /**
      * @return bool
      */
-    public function isBlank()
+    public function isEmpty()
     {
-        return $this->type === self::TYPE_BLANK;
+        return $this->type === self::TYPE_EMPTY;
     }
 
     /**
+     * Not used at the moment
      * @return bool
      */
     public function isFormula()
--- a/src/Spout/Writer/Common/Cell.php
+++ b/src/Spout/Writer/Common/Cell.php
@@ -1,0 +1,122 @@
+<?php
+
+namespace Box\Spout\Writer\Common;
+
+use Box\Spout\Writer\Common\Helper\CellHelper;
+
+class Cell
+{
+    /**
+     * Numeric cell type (whole numbers, fractional numbers, dates)
+     */
+    const CELL_TYPE_NUMERIC = 0;
+
+    /**
+     * String (text) cell type
+     */
+    const CELL_TYPE_STRING = 1;
+
+    /**
+     * Formula cell type
+     */
+    const CELL_TYPE_FORMULA = 2;
+
+    /**
+     * Blank cell type
+     */
+    const CELL_TYPE_BLANK = 3;
+
+    /**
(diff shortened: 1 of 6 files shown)


boxcla:

adrilo:
Thanks @madflow ! From a really quick scan, this is looking good :)
I need to take more time to review it correctly and this might take some time. Apologizing in advance...

davidNieves:
Hi!

With this new functionality is possible set different style each cell in same row? Can you put an example use?

Thanks!!

madflow:
This is just a first step towards this functionality. It is not possible yet.

src/Spout/Writer/ODS/Internal/Worksheet.php
```diff
@@ -191,27 +192,33 @@ protected function getCellXML($cellValue, $styleIndex, $numTimesValueRepeated)
             $data .= ' table:number-columns-repeated="' . $numTimesValueRepeated . '"';
         }
 
-        if (CellHelper::isNonEmptyString($cellValue)) {
+        if ($cellValue instanceof Cell) {
```
adrilo: Can this be moved in the calling function to avoid duplication across all writers?

src/Spout/Writer/Common/Cell.php
```diff
@@ -0,0 +1,160 @@
...
+
+    /**
+     * Formula cell type
+     */
+    const TYPE_FORMULA = 2;
+
+    /**
+     * Blank cell type
+     */
+    const TYPE_BLANK = 3;
```
adrilo: There is now "Blank" and "Empty". If you think "Blank" is a better name, can you rename the isEmpty method from CellHelper to isBlank?
I don't have any preference so feel free to pick whatever you like the best :)

src/Spout/Writer/Common/Cell.php
```diff
@@ -0,0 +1,160 @@
...
+     */
+    const TYPE_NUMERIC = 0;
+
+    /**
+     * String (text) cell type
+     */
+    const TYPE_STRING = 1;
+
+    /**
+     * Formula cell type
```
adrilo: Can you either add a comment saying that this is not used yet or remove it alltogether?

src/Spout/Writer/Common/Cell.php
```diff
@@ -0,0 +1,160 @@
...
+
+    /**
+     * The value of this cell
+     * @var null | mixed
+     */
+    protected $value = null;
+
+    /**
+     * The cell type
+     * @var null
```
adrilo: @var int

src/Spout/Writer/Common/Cell.php
```diff
@@ -0,0 +1,160 @@
...
+     */
+    protected $type = null;
+
+    /**
+     * Cell constructor.
+     * @param $value mixed
+     * @param $comment string
```
adrilo: no `$comment`

src/Spout/Writer/Common/Cell.php
```diff
@@ -0,0 +1,160 @@
...
+    const TYPE_BOOLEAN = 4;
+
+    /**
+     * Error cell type
+     */
+    const TYPE_ERROR = 5;
+
+    /**
+     * The value of this cell
+     * @var null | mixed
```
adrilo: `@var mixed|null` to be consistent

src/Spout/Writer/Common/Cell.php
```diff
@@ -0,0 +1,160 @@
+<?php
+
+namespace Box\Spout\Writer\Common;
+
+use Box\Spout\Writer\Common\Helper\CellHelper;
+
+class Cell
```
adrilo: Add 
```
/**
 * Class Cell
 *
 * @package Box\Spout\Writer\Common
 */
```

adrilo:
Left some minor comments but it's great! I'm excited about finally making some progress in this direction :)

src/Spout/Writer/ODS/Internal/Worksheet.php
madflow: Mmmh. I could move it up in the dedicated writers - but it would still be duplicated. XLSX and ODS do some magic on their own when iterating over the "rows" array. Maybe a future refactoring step?

src/Spout/Writer/Common/Cell.php
madflow: I went with documenting it as "not used"

src/Spout/Writer/Common/Cell.php
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * @return mixed|null
+     */
+    public function getValue()
+    {
+        return $this->value;
+    }
+
+    /**
+     * @return mixed|null
```
adrilo: @return int|null

src/Spout/Writer/Common/Cell.php
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * The value of this cell
+     * @var mixed|null
+     */
+    protected $value = null;
+
+    /**
+     * The cell type
+     * @var int
```
adrilo: int|null

src/Spout/Writer/ODS/Internal/Worksheet.php
adrilo: Ok, we can refactor later. But eventually we should promote the use of `Cell` objects. If it isn't used, the scalar value should be converted to a `Cell` object as early as possible so that all internal components only work with a single type.

Maybe add a TODO so we don't forget about it?

adrilo:
Hey @madflow, I finally had time to go through your changes. I'm sorry it took so long (I should be more responsive now).
Also, what do you think of creating a 3.0/develop branch to merge these changes into?

adrilo:
@madflow I added a `develop_3.0` branch. We can continue the development of these features in this branch.

madflow:
👍  Any chance we can raise the minimum PHP version for the "3.0" branch? I am not missing anything in particular (generators maybe) but it seems like a sane thing to do ;)

adrilo:
That's a good question. We can probably make a case to only support the supported PHP versions (PHP >= 5.6).

Issue #386, opened 2017-02-22
Missing support for getActiveSheet


lijupm:
Libraries like PHPExcel has 'getActiveSheet' method which can get you the active sheet(the sheet which is opened) from your workbook.

It would be better if we have something like the following to get active sheet easily

`$sheet = $reader->getSheetIterator()->getActiveSheet();`

adrilo:
Hi @lijupm, I think that's a good idea.
I left some comments on your pull request as it can't be merged as is.

adrilo:
@lijupm It turns out it is not possible to add a `getActiveSheet()` method for ODS files (or rather no way that would not be incredibly slow). So instead, I implemented a `isActive()` method on the `Sheet` object. You can't get the active sheet directly and still have to use a foreach loop, but you now have a way to know if the current sheet was the active one.

PR #387, opened 2017-02-22
Adding support for 'getActiveSheet' method

lijupm:
This is actually an enhancement - https://github.com/box/spout/issues/386

Libraries like PHPExcel has 'getActiveSheet' method which can get you the active sheet(the sheet which is opened) from your workbook. 

After this change, you can use following code to get the active sheet

`$sheet = $reader->getSheetIterator()->getActiveSheet();`

### DIFF
--- a/src/Spout/Reader/XLSX/Helper/SheetHelper.php
+++ b/src/Spout/Reader/XLSX/Helper/SheetHelper.php
@@ -85,7 +85,7 @@ class SheetHelper
         if ($xmlReader->openFileInZip($this->filePath, self::WORKBOOK_XML_FILE_PATH)) {
             while ($xmlReader->read()) {
                 if ($xmlReader->isPositionedOnStartingNode('workbookView')) {
-                    $activeSheetIndex = $xmlReader->getAttribute('activeTab');
+                    $activeSheetIndex = (int) $xmlReader->getAttribute('activeTab');
                 }
             }
             $xmlReader->close();
--- a/src/Spout/Reader/XLSX/Helper/SheetHelper.php
+++ b/src/Spout/Reader/XLSX/Helper/SheetHelper.php
@@ -73,6 +73,28 @@ class SheetHelper
     }
 
     /**
+     * Get active sheet index
+     *
+     * @return int index of active sheet
+     */
+    public function getActiveSheetIndex()
+    {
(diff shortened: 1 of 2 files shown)


boxcla:
Verified that @lijupm has signed the CLA. Thanks for the pull request!

src/Spout/Reader/XLSX/Helper/SheetHelper.php
```diff
@@ -73,6 +73,28 @@ public function getSheets()
...
+     * Get active sheet index
+     *
+     * @return int index of active sheet
+     */
+    public function getActiveSheetIndex()
+    {
+        $activeSheetIndex = 0;
+
+        $xmlReader = new XMLReader();
+        if ($xmlReader->openFileInZip($this->filePath, self::WORKBOOK_XML_FILE_PATH)) {
```
adrilo: So we basically reading the same file twice. Not good... This should be done in one pass (combining this code with the code from `getSheets`.

src/Spout/Reader/XLSX/SheetIterator.php
```diff
@@ -20,6 +20,9 @@ class SheetIterator implements IteratorInterface
     /** @var int The index of the sheet being read (zero-based) */
     protected $currentSheetIndex;
 
+    /** @var  int The index of the sheet which is active */
+    protected $activeSheetIndex;
```
adrilo: remove extra space between @var and int. Also, it can be initialized here to 0.

src/Spout/Reader/XLSX/SheetIterator.php
```diff
@@ -32,13 +35,24 @@ public function __construct($filePath, $options, $sharedStringsHelper, $globalFu
         // Fetch all available sheets
         $sheetHelper = new SheetHelper($filePath, $options, $sharedStringsHelper, $globalFunctionsHelper);
         $this->sheets = $sheetHelper->getSheets();
+        $this->activeSheetIndex = $sheetHelper->getActiveSheetIndex();
```
adrilo: this should be done in `getSheets`directly

adrilo:
Can you please add some tests? And ideally, the same behavior should be added for ODS files and the documentation be updated :)

adrilo:
Fixed as part of #405

PR #395, opened 2017-03-21
Fix table

garak:
Looks like it's a bug in markdown parser. Null blank space is a bit ugly, but it does the trick


boxcla:
Verified that @garak has signed the CLA. Thanks for the pull request!

adrilo:
Better fix here: #398. Thanks for pointing this out @garak !

PR #397, opened 2017-03-27
Enforce sheet name uniqueness per workbook

### LINKED ISSUE #391: InvalidSheetNameException: The sheet ("blabla") is invalid. It did not respect these rules: - It should be unique ...
I was using this class to open and process a number of excel files in a loop and got the message :

ox\Spout\Writer\Exception\InvalidSheetNameException: The sheet&#039;s name ("blabla") is invalid. It did not respect these rules:
 - It should be unique in Box\Spout\Writer\Common\Sheet->throwIfNameIsInvalid() (line 128 of D:\wamp\www\sites\all\libraries\spout-master\src\Spout\Writer\Common\Sheet.php).

By debugging the code I noticed that somehow the class was referring to sheetnames that were used in a previous processing (even though I cleared everything before reusing the class again).

I found that the file "spout-master\src\Spout\Writer\Common\Sheet.php" did not initialize the static variable "$SHEETS_NAME_USED" and as result the names in there show up again when you reuse the class.

So I changed (added the line in bold) the following function (in the same file) to resolve the issue:

```
public function __construct($sheetIndex)
    {
        **self::$SHEETS_NAME_USED = array();**
		$this->index = $sheetIndex;
        $this->stringHelper = new StringHelper();
        $this->setName(self::DEFAULT_SHEET_NAME_PREFIX . ($sheetIndex + 1));
		
    }
```

Kind regards,

Jos.

adrilo:
Fixes #391

Instead of checking across all workbooks, enforcement should be done per workbook (in case of multiple spreadsheets being created at the same time).

### DIFF
--- a/src/Spout/Writer/Common/Internal/AbstractWorkbook.php
+++ b/src/Spout/Writer/Common/Internal/AbstractWorkbook.php
@@ -16,6 +16,9 @@ abstract class AbstractWorkbook implements WorkbookInterface
     /** @var bool Whether new sheets should be automatically created when the max rows limit per sheet is reached */
     protected $shouldCreateNewSheetsAutomatically;
 
+    /** @var string Timestamp based unique ID identifying the workbook */
+    protected $internalId;
+
     /** @var WorksheetInterface[] Array containing the workbook's sheets */
     protected $worksheets = [];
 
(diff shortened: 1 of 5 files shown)


boxcla:

PR #401, opened 2017-03-28
Do not add space between text nodes

### LINKED ISSUE #400: reading cell value incorrectly
I have a excel file (https://github.com/box/spout/files/875118/space.problem.xlsx). when I using Reader to get the text value of cell, it added an space in the result

I extract my excel file, I found that my value are stored in sharedStrings.xml

`       <si>
		<r>
			<t>t</t>
		</r><r>
			<rPr>
				<sz val="12"/>
				<color rgb="FF000000"/>
				<rFont val="Calibri"/>
				<family val="2"/>
			</rPr><t>est</t>
		</r>
		<phoneticPr fontId="1" type="noConversion"/>
	</si>`

I alse found the Reader/XLSX/Helper/SharedStringsHelper.php line 239 $textValue .= ' ';
I wondering why add a space between each "t" node

adrilo:
Fixes #400 

It was added to satisfy a test but the test file was missing the `xml:space="preserve"` attribute to get the expected result. This commit fixes the code and the test file.


boxcla:

PR #402, opened 2017-03-28
Prevent error when close() called while writer already closed

### LINKED ISSUE #392: Xlsx writer closes file even if already closed.
Hi,

Encountered an issue when trying to call close() method twice:
> [Box\Spout\Common\Exception\IOException]                       
>   Cannot perform I/O operation outside of the base folder: /tmp

It seems that writer checks for book to be present, even if isWriterOpened flag is false:
```
/**
     * Closes the writer, preventing any additional writing.
     *
     * @return void
     */
    protected function closeWriter()
    {
        if ($this->book) {
            $this->book->close($this->filePointer);
        }
    }
```

I don't have time to prepare PR now, so I suggest adding  `&& $this->isWriterOpened` condition in 
[src/Spout/Writer/XLSX/Writer.php:128](https://github.com/box/spout/blob/master/src/Spout/Writer/XLSX/Writer.php#L128).

Thanks in advance!
P.S. Let me know if you need a PR for this, I'll do my best to create one in a few days.
P.P.S. I tried to implement some cleanup / rollback functionality. I suppose creating a public method with `closeAndAttemptToCleanupAllFiles()` code inside would be nice ;)

adrilo:
Fixes #392 

Before trying to close the writer, check that it's not already closed.

### DIFF
--- a/src/Spout/Writer/AbstractWriter.php
+++ b/src/Spout/Writer/AbstractWriter.php
@@ -350,6 +350,10 @@ abstract class AbstractWriter implements WriterInterface
      */
     public function close()
     {
+        if (!$this->isWriterOpened) {
+            return;
+        }
+
         $this->closeWriter();
 
         if (is_resource($this->filePointer)) {
@@ -378,4 +382,3 @@ abstract class AbstractWriter implements WriterInterface
         }
     }
 }
-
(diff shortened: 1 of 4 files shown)


boxcla:

PR #405, opened 2017-04-15
Expose Sheet::isActive() to provide info about the last active sheet

### LINKED ISSUE #386: Missing support for getActiveSheet

adrilo:
Fixes #386

The API is exposed for all types of readers.


boxcla:

PR #411, opened 2017-04-28
Fix shared strings XML Entities auto decode

### LINKED ISSUE #409: issue : In box spout,when reading cell with "&#Number" content but that content  "&#Number" is not shown.
Hi,
When I try to read the cell values like "03-0129-1-HP LED 18.5& #34;19KA Monitor" using box spout.It writtens only "03-0129-1-HP LED 18.5''19KA Monitor" this value. The "&#34;" content is eliminated.
What Should I do?

In the above question don't use space between & and #. why I am using space is,in github without space between & and # is converted into blank.

adrilo:
Fix #409 

When converting an XMLReader node to a SimpleXMLElement, the conversion would automatically decode the XML entities. This resulted in a double decode.
For example: "& amp;#34;" was converted to "& #34;" when imported into a SimpleXMLElement and was again converted into " (quote).

This commit changes the way the XLSX Shared Strings file is processed. It also changes the unescaping logic for both XLSX and ODS.

Finally, it removes any usage of the SimpleXML library (yay!).


boxcla:

timhunt:
Hmm. This 'Fix' caused a regression #576.

"test" (including the double-quotes) is a valid worksheet name in Excel, and worksheet names are stored in XML attributes. Therefore it is vital to escape them.

The commit for this issue changed everything to ENT_NOQUOTES, with comments to say do that, and unit tests to verify that. However, it does not say whether that is really important for any reason, or whether you just thought it looked nicer that way.

I got here from git history, trying to understand whether it is safe to change it back to ENT_QUOTES, but I am still not sure if it is.

Thanks in advance if you can give me any help.

PR #418, opened 2017-05-19
Make XLSX shared strings path case insensitive

11ya:
Some XLSX files have shared strings XML name started with uppercase letter (SharedStrings.xml). So XMLReader can't find required file in that situation.

### DIFF
--- a/src/Spout/Reader/XLSX/Helper/SharedStringsHelper.php
+++ b/src/Spout/Reader/XLSX/Helper/SharedStringsHelper.php
@@ -60,15 +60,28 @@ class SharedStringsHelper
      */
     public function hasSharedStrings()
     {
-        $hasSharedStrings = false;
+        return $this->getSharedStringsPath() !== null;
+    }
+
+    /**
+     * Returns shared strings XML file path
+     *
+     * @return string|null
+     */
+    public function getSharedStringsPath()
+    {
+        $sharedStringsPath = null;
         $zip = new \ZipArchive();
 
         if ($zip->open($this->filePath) === true) {
-            $hasSharedStrings = ($zip->locateName(self::SHARED_STRINGS_XML_FILE_PATH) !== false);
+            $index = $zip->locateName(self::SHARED_STRINGS_XML_FILE_PATH, \ZipArchive::FL_NOCASE);
+            if ($index !== false) {
+                $sharedStringsPath = $zip->getNameIndex($index);
+            }
             $zip->close();
         }
 
-        return $hasSharedStrings;
+        return $sharedStringsPath;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


boxcla:
Hi @11ya, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

11ya:
CLA signed

boxcla:
Verified that @11ya has just signed the CLA. Thanks, and we look forward to your contribution.

src/Spout/Reader/XLSX/Reader.php
```diff
@@ -74,9 +74,9 @@ protected function openReader($filePath)
         if ($this->zip->open($filePath) === true) {
             $this->sharedStringsHelper = new SharedStringsHelper($filePath, $this->getOptions()->getTempFolder());
 
-            if ($this->sharedStringsHelper->hasSharedStrings()) {
+            if ($sharedStringsPath = $this->sharedStringsHelper->getSharedStringsPath()) {
```
adrilo: The shared strings path should remain private. There's no need to expose the path: the path should only be known by the shared strings helper. Therefore, I would keep the `hasSharedStrings()` call here and don't pass anything to the `extractSharedStrings`

src/Spout/Reader/XLSX/Helper/SharedStringsHelper.php
```diff
@@ -60,15 +60,28 @@ public function __construct($filePath, $tempFolder = null)
...
-        $hasSharedStrings = false;
+        return $this->getSharedStringsPath() !== null;
+    }
+
+    /**
+     * Returns shared strings XML file path
+     *
+     * @return string|null
+     */
+    public function getSharedStringsPath()
```
adrilo: This function's name is confusing, as there already is a `getSharedStringsFilePath()` function in this file

src/Spout/Reader/XLSX/Helper/SharedStringsHelper.php
```diff
@@ -60,15 +60,28 @@ public function __construct($filePath, $tempFolder = null)
...
+    {
+        $sharedStringsPath = null;
         $zip = new \ZipArchive();
 
         if ($zip->open($this->filePath) === true) {
-            $hasSharedStrings = ($zip->locateName(self::SHARED_STRINGS_XML_FILE_PATH) !== false);
+            $index = $zip->locateName(self::SHARED_STRINGS_XML_FILE_PATH, \ZipArchive::FL_NOCASE);
```
adrilo: instead of using a constant and try to find the correct file name by locating it inside the zip file, we should probably get it from the `[Content_Types].xml` file.

It looks something like this:
```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types">
    ....
    <Override PartName="/xl/sharedStrings.xml" ContentType="application/vnd.openxmlformats-officedocument.spreadsheetml.sharedStrings+xml"/>
</Types>
```

adrilo: This seems a more robust approach (in case, someone decides to change the name of the file alltogether)

11ya:
Closing to move to separate branch and resolve conflicts

PR #419, opened 2017-05-21
bump phpunit to the latest supported release, travis setup, ignore co…

madflow:
* bump phpunit to the latest supported release 5.7 that supports PHP 5.6 (EOL is 2018-02) 
* travis setup
* ignore composer.lock (should not be part of libraries)


boxcla:

adrilo:
To fix Scrutinizer, you may need to update .scrutinizer.yml (the after_script section)

madflow:
I pinned the coverage to only send data in PHP 7.1 and rebased. However - the Scrutinizer job took unusually long. This could be just a coincidence (worker queue full or sth...) I guess... I can monitor this in the next commits.

adrilo:
Indeed, it took forever !! Right, we'll see if that's only a temporary issue.

PR #423, opened 2017-05-24
Introduce an options manager for all writers

adrilo:
This will improve the management of options and simplify some methods' signatures.
This commit will also help moving the code to a better state regarding Dependency Injection.


boxcla:

Issue #424, opened 2017-05-25
Write-Only XLSX formulas.


stevesweets:
I've actually done this and was about to submit a pull request when i noticed, that actually, one should file an issue first.

Reading them uses memory. Writing them is really performant.

adrilo:
No plans to support formulas. Closing this issue.

PR #352, opened 2016-11-02
Excel export: added a special case to add HYPERLINK Formula when text…

mdsimpson:
… in the cell matches =HYPERLINK("link","name")

A very basic start at adding support for Excel function.

PR #425, opened 2017-05-25
Write formulas to XLSX files.

stevesweets:
Ref this issue: https://github.com/box/spout/issues/424

I utterly understand that reading formulas is a pain.  It should be noted that writing formulas to ODS is rather a pain, too, but the nature of xlsx actually makes it rather easy to write them.

This way, we can use the super-performant spout when we want to generate an xlsx file that needs to contain some excel formulas, and I don't think this lands outside of the goals of Spout.

I await your judgement!

### DIFF
--- a/src/Spout/Writer/XLSX/Internal/Worksheet.php
+++ b/src/Spout/Writer/XLSX/Internal/Worksheet.php
@@ -212,8 +212,10 @@ EOD;
         $columnIndex = CellHelper::getCellIndexFromColumnIndex($cellNumber);
         $cellXML = '<c r="' . $columnIndex . $rowIndex . '"';
         $cellXML .= ' s="' . $styleId . '"';
-
-        if (CellHelper::isNonEmptyString($cellValue)) {
+        
+        if (CellHelper::isFormulaString($cellValue)) {
+            $cellXML .= '><f>'.substr($cellValue,1).'</f><v>0</v></c>'; // seriously, that's it.
+        } else if (CellHelper::isNonEmptyString($cellValue)) {
             $cellXML .= $this->getCellXMLFragmentForNonEmptyString($cellValue);
         } else if (CellHelper::isBoolean($cellValue)) {
             $cellXML .= ' t="b"><v>' . intval($cellValue) . '</v></c>';
--- a/src/Spout/Writer/Common/Helper/CellHelper.php
+++ b/src/Spout/Writer/Common/Helper/CellHelper.php
@@ -65,6 +65,17 @@ class CellHelper
     }
 
     /**
+     * Returns whether the given value looks like a formula
+     *
+     * @param $value
+     * @return bool whether the given value looks like a formula
+     */
+    public static function isFormulaString($value)
+    {
+        return (strpos($value,'=') === 0);
+    }
+
+    /**
      * Returns whether the given value is numeric.
      * A numeric value is from type "integer" or "double" ("float" is not returned by gettype).
      *


boxcla:
Verified that @stevesweets has signed the CLA. Thanks for the pull request!

stevesweets:
This works for this, too:

https://github.com/box/spout/pull/352

src/Spout/Writer/XLSX/Internal/Worksheet.php
```diff
@@ -212,8 +212,10 @@ private function getCellXML($rowIndex, $cellNumber, $cellValue, $styleId)
         $columnIndex = CellHelper::getCellIndexFromColumnIndex($cellNumber);
         $cellXML = '<c r="' . $columnIndex . $rowIndex . '"';
         $cellXML .= ' s="' . $styleId . '"';
-
-        if (CellHelper::isNonEmptyString($cellValue)) {
+        
+        if (CellHelper::isFormulaString($cellValue)) {
+            $cellXML .= '><f>'.substr($cellValue,1).'</f><v>0</v></c>'; // seriously, that's it.
```
adrilo: I suspect that the `<v>0</v>` is not going to work across all softwares. Which software did you use to test your changes?

adrilo: Also, by setting the value to `0`, reading back a spreadsheet with a formula won't work as Spout would return "0" for the cell value. Not ideal...

adrilo:
Thanks @stevesweets !
Most likely, this won't be merged until v3 is released. v3 should make it easier to work with formulas. But I'll keep the PR open, as a reference for future implementation.

src/Spout/Writer/XLSX/Internal/Worksheet.php
stevesweets: Tested on OSX, with excel 15 and numbers.

<v>0</v> works because excel formulas are recalculated when an excel file is loaded.

You are correct that _reading_ a formula will result in unexpected results, I had hoped the title of the PR was specific enough for this to be clear.

Thanks for coming back to me, glad to have been of any use at all.

SimZal: I tried it on Win10 and LibreOffice and it doesn't seem to work. It displays 0, like adrilo predicted.

adrilo: @stevesweets can you try the empty `<v></v>`?
There may always be some software that won't work but if the main ones work, we can consider merging this change.

stevesweets: Hi there, realise i left this a long time - i will try this.

S

Yursksf1: how is the code for .ods?

Yursksf1: I try do the change on code for .ods 
[ODS/Internal/Worksheet.php](https://github.com/box/spout/blob/master/src/Spout/Writer/ODS/Internal/Worksheet.php)
```php
        if (CellHelper::isFormulaString($cellValue)) {
            $data .= ' table:formula="'.$cellValue.'" office:value-type="float" office:value="' . 0 . '" calcext:value-type="float">';
            $data .= '<text:p>' . 0 . '</text:p>';
        }else if (CellHelper::isNonEmptyString($cellValue)) {
            $data .= ' office:value-type="string" calcext:value-type="string">';
            $cellValueLines = explode("\n", $cellValue);
            foreach ($cellValueLines as $cellValueLine) {
                $data .= '<text:p>' . $this->stringsEscaper->escape($cellValueLine) . '</text:p>';
            }
        } else if (CellHelper::isBoolean($cellValue)) {...}
```
too add the same change the [CellHelper.php](https://github.com/box/spout/blob/master/src/Spout/Writer/Common/Helper/CellHelper.php) like .xlsx

```php
 public static function isFormulaString($value)
    {
        return (strpos($value,'=') === 0);
    }
```
but now have this problem
not reconize the form in ods like excel. i neet to open document "make change at formule" and save to func 
image
image

this is the code .fods
```php
 <table:table-row table:style-name="ro1">
     <table:table-cell office:value-type="float" office:value="1" calcext:value-type="float">
      <text:p>1</text:p>
     </table:table-cell>
     <table:table-cell office:value-type="float" office:value="9" calcext:value-type="float">
      <text:p>9</text:p>
     </table:table-cell>
     <table:table-cell office:value-type="float" office:value="3" calcext:value-type="float">
      <text:p>3</text:p>
     </table:table-cell>
     <table:table-cell office:value-type="float" office:value="7" calcext:value-type="float">
      <text:p>7</text:p>
     </table:table-cell>
     <table:table-cell table:formula="of:=[.A1]+[.B1]+[.C1]+[.D1]" office:value-type="float" office:value="20" calcext:value-type="float">
      <text:p>20</text:p>
     </table:table-cell>
    </table:table-row>
```

if any can help me i give thanks

adrilo: Well, you can see that formulas don't work the same:
```
<table:table-cell table:formula="of:=[.A1]+[.B1]+[.C1]+[.D1]" office:value-type="float" office:value="20" calcext:value-type="float">
      <text:p>20</text:p>
```
This would indicate a formula: `table:formula="of:=[.A1]+[.B1]+[.C1]+[.D1]"`, so your `isFormulaString` implementation is incorrect. Setting the value to 0 may not work in ODS. This is a trick in XLSX because softwares do recompute the value when you open the file, but for ODS maybe softwares use the given value directly. In this case, I don't have a good solution for you.

StefanoGroenland:
Will this PR ever get merged?

Slamdunk:
Merged a variation in https://github.com/openspout/openspout/pull/5

PR #428, opened 2017-05-29
Move OptionManager from Common/Manager to Manager

adrilo:


boxcla:

PR #433, opened 2017-05-30
Refactory Writer Styles to match new code organization

adrilo:
Decomposed old StyleHelper into StyleManager, StyleRegistry and StyleMerger.


boxcla:

PR #434, opened 2017-05-30
Row objects and Cell styling

madflow:
This PR introduces a ```Row``` object in all writers and implements ```Cell``` styling for #182.

I left the public writer methods

```
WriterAbstract::addRow
WriterAbstract::addRows
WriterAbstract::addRowWithStyle
WriterAbstract::addRowsWithStyle
```

intact - so all the existing tests can pass. However the signatured changed. This can be considered a BC Break. Because the legacy data arrays are transformed into Row objects - it is hacky. When we should drop the support for arrays - it should get better.

* A ```Row``` object holds the cells and styles of a row.
* From that point on (after the Writer) - all signatures like```(array $dataRows, $style)``` have been refactored to ```(Row $row)```.
* With the newly created ```Row``` object the *WithStyle mehods can be marked as @deprecated or can be deleted.
* This also made managing the row style in the ```Writer``` superfluous.
* I added a WriterAbstract::withRow method that accepts a closure.
* Registering the styles is now done right before a ```Cell``` is written to the output.
* The default style is merged into the row style - and the row style is merged into the cell style.
* The tests pass - more can follow when we agree that this is the "right" way to proceed.

Example:

```PHP
use Box\Spout\Common\Type;
use Box\Spout\Writer\Common\Creator\Style\BorderBuilder;
use Box\Spout\Writer\Common\Creator\Style\StyleBuilder;
use Box\Spout\Writer\Common\Entity\Cell;
use Box\Spout\Writer\Common\Entity\Row;
use Box\Spout\Writer\Common\Entity\Style\Color;
use Box\Spout\Writer\WriterFactory;


$defaultStyleBorderBottom = (new StyleBuilder())
    ->setBorder((new BorderBuilder())->setBorderBottom()->build())
    ->build();

$rowStyleItalicBold = (new StyleBuilder())
    ->setFontItalic()
    ->setFontBold()
    ->build();

$greenLight = (new StyleBuilder())
    ->setBackgroundColor(Color::GREEN)
    ->build();

$yellowLight = (new StyleBuilder())
    ->setBackgroundColor(Color::YELLOW)
    ->build();

$redLight = (new StyleBuilder())
    ->setBackgroundColor(Color::RED)
    ->build();

/** @var \Box\Spout\Writer\XLSX\Writer $writer */
$writer = WriterFactory::create(Type::XLSX);
$writer->setDefaultRowStyle($defaultStyleBorderBottom);
$writer->openToFile(__DIR__ . '/test.xlsx');

$row = new Row();
$row
    ->addCell(new Cell('G', $greenLight))
    ->addCell(new Cell('Y', $yellowLight))
    ->addCell(new Cell('R', $redLight))
    ->setStyle($rowStyleItalicBold);

$writer->addRow($row);

$writer->close();
```

This yields:

screen

or:

```PHP

$writer->withRow(function(Row $row) use ($greenLight, $yellowLight, $redLight, $rowStyleItalicBold) {
    return $row
        ->addCell(new Cell('G', $greenLight))
        ->addCell(new Cell('Y', $yellowLight))
        ->addCell(new Cell('R', $redLight))
        ->setStyle($rowStyleItalicBold);
});
```


boxcla:

adrilo:
Thanks @madflow !

I thought about introducing a `Row` object but was not sure about it... I was afraid that it would make it harder to get started and feel too Java-ish. Something we can do though is to introduce a method that would convert an array to a row of simple cells?

I'm gonna take a look at the code soon. BTW I'm refactoring most of the code at the moment and you may need to move code around as I commit new changes. Sorry about that...

madflow:
There already is a protected AbstractWriter::createRowFromArray method that does this. Something similar can be made public in some sort of way.

I agree that the ```Row```adds some bloat  - but when it comes to cell styling - I wanted a fluent interface  for adding cells. 

I recognized your commits (the hard way :D ) - the last one to merge looses I guess...

src/Spout/Writer/Common/Entity/Cell.php
```diff
@@ -81,6 +100,25 @@ public function getValue()
     }
 
     /**
+     * @param Style $style
```
adrilo: Style|null

src/Spout/Writer/Common/Entity/Cell.php
```diff
@@ -55,12 +57,29 @@ class Cell
...
+    /**
      * Cell constructor.
      * @param $value mixed
+     * @param $style Style
      */
-    public function __construct($value)
+    public function __construct($value, Style $style = null)
     {
         $this->setValue($value);
+        if ($style) {
```
adrilo: You can probably remove the `if` statement, as it's fine to set the style to `null`

src/Spout/Writer/Common/Entity/Cell.php
```diff
@@ -81,6 +100,25 @@ public function getValue()
...
+    {
+        $this->style = $style;
+    }
+
+    /**
+     * @return Style
+     */
+    public function getStyle()
+    {
+        if (!isset($this->style)) {
```
adrilo: It looks wrong to have `getStyle` call `setStyle`. What about setting `$this->style` to `new Style()` when `setStyle(null)` is called (like in the constructor)? This way your getter will simply be a return statement

src/Spout/Writer/Common/Entity/Cell.php
```diff
@@ -164,4 +202,18 @@ public function __toString()
     {
         return (string)$this->value;
     }
+
+    /**
+     * @param Style $style|null
+     * @return $this
+     */
+    public function applyStyle(Style $style = null)
```
adrilo: This should be moved to a CellManager: `CellManager::applyStyle(Cell $cell, Style $style = null)`
The idea is to keep entities logic-less, as much as possible. The logic should live in (as much as possible) stateless managers.

src/Spout/Writer/Common/Entity/Cell.php
```diff
@@ -55,12 +57,29 @@ class Cell
...
      */
-    public function __construct($value)
+    public function __construct($value, Style $style = null)
     {
         $this->setValue($value);
+        if ($style) {
+            $this->setStyle($style);
+        }
+
+        $this->styleMerger = new StyleMerger();
```
adrilo: This should be injected through a factory. See comment below about CellManager.
You can instantiate a new Cell using the `EntityFactory` and the CellManager using the `ManagerFactory`

src/Spout/Writer/Common/Entity/Row.php
```diff
@@ -0,0 +1,117 @@
...
+     * @param array $cells
+     * @param Style|null $style
+     */
+    public function __construct(array $cells = [], Style $style = null)
+    {
+        $this
+            ->setCells($cells)
+            ->setStyle($style);
+
+        $this->styleMerger = new StyleMerger();
```
adrilo: same for the injection

src/Spout/Writer/ODS/Manager/WorksheetManager.php
```diff
@@ -25,18 +25,23 @@ class WorksheetManager implements WorksheetManagerInterface
...
      * @param StringHelper $stringHelper
      */
     public function __construct(
+        StyleManager $styleManager,
         \Box\Spout\Common\Escaper\ODS $stringsEscaper,
         StringHelper $stringHelper)
     {
         $this->stringsEscaper = $stringsEscaper;
         $this->stringHelper = $stringHelper;
+        $this->styleManager = $styleManager;
```
adrilo: [Minor] can you move this to the 1st line to match the order of the arguments?

src/Spout/Writer/ODS/Manager/WorksheetManager.php
```diff
@@ -87,40 +92,44 @@ public function getTableElementStartAsString(Worksheet $worksheet)
     }
 
     /**
-     * Adds data to the given worksheet.
+    /**
```
adrilo: remove

src/Spout/Writer/ODS/Manager/WorksheetManager.php
```diff
@@ -87,40 +92,44 @@ public function getTableElementStartAsString(Worksheet $worksheet)
...
         $nextCellIndex = 1;
 
         for ($i = 0; $i < $cellsCount; $i++) {
-            $currentCellValue = $dataRowWithNumericIndexes[$currentCellIndex];
 
-            // Using isset here because it is way faster than array_key_exists...
-            if (!isset($dataRowWithNumericIndexes[$nextCellIndex]) ||
-                $currentCellValue !== $dataRowWithNumericIndexes[$nextCellIndex]) {
+            /** @var Cell $cell */
+            $cell = $row->getCells()[$currentCellIndex];
```
adrilo: $cell = $cells[$currentCellIndex]

adrilo: or `$cell = $row->getCellAtIndex($currentCellIndex);` ?

src/Spout/Writer/ODS/Manager/WorksheetManager.php
```diff
@@ -87,40 +92,44 @@ public function getTableElementStartAsString(Worksheet $worksheet)
...
+            /** @var Cell|null $nextCell */
+            $nextCell = isset($cells[$nextCellIndex]) ? $cells[$nextCellIndex] : null;
 
-                $numTimesValueRepeated = ($nextCellIndex - $currentCellIndex);
-                $data .= $this->getCellXML($currentCellValue, $styleIndex, $numTimesValueRepeated);
+            if (null === $nextCell || $cell->getValue() !== $nextCell->getValue()) {
+
+                // Apply styles - the row style is merged at this point
+                $cell->applyStyle($row->getStyle());
```
adrilo: move this block to its own function

src/Spout/Writer/ODS/Manager/WorksheetManager.php
```diff
@@ -142,27 +151,20 @@ public function addRow(Worksheet $worksheet, $dataRow, $rowStyle)
...
-    private function getCellXML($cellValue, $styleIndex, $numTimesValueRepeated)
+    protected function getCellXML(Cell $cell, $styleIndex, $numTimesValueRepeated)
     {
         $data = '<table:table-cell table:style-name="ce' . $styleIndex . '"';
 
         if ($numTimesValueRepeated !== 1) {
             $data .= ' table:number-columns-repeated="' . $numTimesValueRepeated . '"';
         }
 
-        /** @TODO Remove code duplication with XLSX writer: https://github.com/box/spout/pull/383#discussion_r113292746 */
```
adrilo: 👍

src/Spout/Writer/WriterAbstract.php
```diff
@@ -199,64 +193,111 @@ protected function throwIfWriterAlreadyOpened($message)
...
-     * @return WriterAbstract
-     * @throws \Box\Spout\Writer\Exception\WriterNotOpenedException If this function is called before opening the writer
-     * @throws \Box\Spout\Common\Exception\IOException If unable to write data
-     * @throws \Box\Spout\Common\Exception\SpoutException If anything else goes wrong while writing data
      */
-    public function addRow(array $dataRow)
+    public function addRow($row)
     {
+        if (!is_array($row) && !$row instanceof Row) {
+            throw new InvalidArgumentException('addRow accepts an array with scalar values or a Row object');
```
adrilo: update phpdoc

src/Spout/Writer/WriterAbstract.php
```diff
@@ -199,64 +193,111 @@ protected function throwIfWriterAlreadyOpened($message)
...
-     * @api
-     * @param array $dataRow Array of array containing data to be streamed.
+     * @param array|\Box\Spout\Writer\Common\Entity\Row $row The row to be appended to the stream
      * @param Style $style Style to be applied to the row.
-     * @return WriterAbstract
-     * @throws \Box\Spout\Common\Exception\InvalidArgumentException If the input param is not valid
-     * @throws \Box\Spout\Writer\Exception\WriterNotOpenedException If this function is called before opening the writer
-     * @throws \Box\Spout\Common\Exception\IOException If unable to write data
+     * @return WriterInterface
+     * @internal param array $row Array containing data to be streamed.
```
adrilo: You can drop support for that anytime, as 3.0 will contain other breaking changes.

src/Spout/Writer/WriterAbstract.php
```diff
@@ -305,35 +344,32 @@ public function addRowsWithStyle(array $dataRows, $style)
             throw new InvalidArgumentException('The "$style" argument must be a Style instance and cannot be NULL.');
         }
 
-        $this->setRowStyle($style);
-        $this->addRows($dataRows);
-        $this->resetRowStyleToDefault();
+        $this->addRows(array_map(function ($row) use ($style) {
```
adrilo: We can probably do better than that, as the row is traversed twice here. Maybe just call `addRow` in a foreach loop? as style is not handled the same way, that may work

src/Spout/Writer/WriterAbstract.php
```diff
@@ -305,35 +344,32 @@ public function addRowsWithStyle(array $dataRows, $style)
...
-
-    /**
-     * Resets the style to be applied to the next written rows.
-     *
-     * @return void
-     */
-    private function resetRowStyleToDefault()
-    {
-        $this->rowStyle = $this->optionsManager->getOption(Options::DEFAULT_ROW_STYLE);
+        if (null === $defaultRowStyle) {
```
adrilo: I would move this check into the `merge` function of styleMerger

src/Spout/Writer/WriterAbstract.php
```diff
@@ -305,35 +344,32 @@ public function addRowsWithStyle(array $dataRows, $style)
...
+            return $this;
+        }
+        $merged = $this->styleMerger->merge($row->getStyle(), $defaultRowStyle);
+        $row->setStyle($merged);
+        return $this;
```
adrilo: No need to return something here

src/Spout/Writer/WriterInterface.php
```diff
@@ -34,26 +34,38 @@ public function openToBrowser($outputFileName);
...
+    /**
+     * Write given data to the output with a closure funtion. New data will be appended to end of stream.
+     *
+     * @param \Closure $callback A callback returning a Row object. A new Row object is injected into the callback.
+     * @return WriterInterface
+     * @internal param \Closure $callback
+     *          Example withRow(function(Row $row) { return $row->addCell('data1'); })
      */
-    public function addRowWithStyle(array $dataRow, $style);
+    public function withRow(\Closure $callback);
```
adrilo: I'm still unsure if this function is a good idea... Having multiple ways to do something is not necessarily better

src/Spout/Writer/XLSX/Manager/WorksheetManager.php
```diff
@@ -115,60 +115,47 @@ private function throwIfSheetFilePointerIsNotAvailable($sheetFilePointer)
...
+        $numCells = count($row->getCells());
 
         $rowXML = '<row r="' . $rowIndex . '" spans="1:' . $numCells . '">';
 
-        foreach($dataRow as $cellValue) {
-            $rowXML .= $this->getCellXML($rowIndex, $cellNumber, $cellValue, $style->getId());
+        /** @var Cell $cell */
+        foreach($row->getCells() as $cell) {
+            // Apply styles - the row style is merged at this point
+            $cell->applyStyle($row->getStyle());
```
adrilo: this can also be moved to its own function

src/Spout/Writer/Common/Manager/WorkbookManagerAbstract.php
```diff
@@ -238,19 +236,16 @@ private function hasCurrentWorkseetReachedMaxRows()
      * @param Worksheet $worksheet Worksheet to write the row to
      * @param array $dataRow Array containing data to be written. Cannot be empty.
      *          Example $dataRow = ['data1', 1234, null, '', 'data5'];
-     * @param Style $style Style to be applied to the row.
      * @return void
      * @throws WriterException If unable to write data
      */
-    private function addRowWithStyleToWorksheet(Worksheet $worksheet, $dataRow, Style $style)
+    private function addRowWithStyleToWorksheet(Worksheet $worksheet, Row $row)
```
adrilo: you can rename addRowToWorksheet

src/Spout/Writer/Common/Entity/Row.php
```diff
@@ -0,0 +1,117 @@
...
+     * @var StyleMerger
+     */
+    protected $styleMerger;
+
+    /**
+     * Row constructor.
+     * @param array $cells
+     * @param Style|null $style
+     */
+    public function __construct(array $cells = [], Style $style = null)
```
adrilo: replace `array $cells` by `Cell[] $cells`

src/Spout/Writer/Common/Entity/Row.php
```diff
@@ -0,0 +1,117 @@
...
+    protected $style = null;
+
+    /**
+     * @var StyleMerger
+     */
+    protected $styleMerger;
+
+    /**
+     * Row constructor.
+     * @param array $cells
```
adrilo: replace `array $cells` by `Cell[] $cells`

src/Spout/Writer/Common/Entity/Cell.php
```diff
@@ -55,12 +57,29 @@ class Cell
...
+    protected $style = null;
+
+    /**
+     * @var StyleMerger
+     */
+    protected $styleMerger;
+
+    /**
      * Cell constructor.
      * @param $value mixed
```
adrilo:
```
@param mixed $value
@param Style|null $style
```

src/Spout/Writer/Common/Entity/Row.php
adrilo: Also, I would make the `$cells` param a required param. If you want an empty row, then just pass an empty array yourself.

Regarding the 2nd argument, should we just allow setting the row style using `Row::setStyle()`? Or is it better to match the `Cell` signature?

adrilo:
@madflow, this looks really good ! I left a bunch of comments but I like the approach.

I think you can drop the support for arrays (in favor of `Row`). If we do that, `addRow` and `addRows` functions won't be backwards compatible. So it will break the code of most users. If code is broken, you can definitely the `*WithStyle` functions as it will break more stuff but since ppl need to update their code to upgrade anyways...

Also, instead of doing `(new Row())->addCell(...)`, you can introduce a RowFactory (+) or RowBuilder (-). This will enable you to inject the dependencies of `Row` (`StyleMerger`).

adrilo:
BTW, I'm done refactoring the Writer code

madflow:
Thanks for the review! Good to know about the status of the refactoring. I'll take a look at the comments and post back.

src/Spout/Writer/Common/Entity/Cell.php
madflow: I did this so there can be unstyled cells. Only when a writer / worksheet / manager requests a style it should be created. This will probably change though with the Cell Manager concept...

src/Spout/Writer/Common/Entity/Row.php
madflow: First I thought of this as a "Row Builder" not an entity. Hence the example without constructure params. But - I'd personally be fine with such a signature. I just like the flexibility here. But - when we add a RowBuilder  - this can change. I'd suggest ```(array $cells, Style $style = null)```

src/Spout/Writer/Common/Manager/CellManager.php
```diff
@@ -0,0 +1,37 @@
...
+     * Merges a Style into a cells Style.
+     *
+     * @param Cell $cell
+     * @param Style $style
+     * @return $this
+     */
+    public function applyStyle(Cell $cell, Style $style)
+    {
+        $mergedStyle = $this->styleMerger->merge($cell->getStyle(), $style);
+        $cell->setStyle($mergedStyle);
```
adrilo: return $cell ? or update the phpdoc `@return void`

src/Spout/Writer/WriterAbstract.php
```diff
@@ -344,20 +344,25 @@ public function addRowsWithStyle(array $dataRows, $style)
...
-        }, $dataRows));
+
+            $this->addRow($row);
+        }
 
         return $this;
     }
 
     /**
+     * @TODO: Move this into styleMerger
```
adrilo: I meant only move the `if (null === $defaultRowStyle)` part. The `merge` function should handle null param

src/Spout/Writer/Common/Entity/Row.php
adrilo: Then to be consistent with the naming of the other objects, you can actually rename that to RowBuilder and add a `build` function that returns the row with the added cells and correct style.

```
$row = new Row();
$row
    ->addCell(new Cell('G', $greenLight))
    ->setStyle($rowStyleItalicBold);
```
would become
```
$row = (new RowBuilder())
    ->addCell(new Cell('G', $greenLight))
    ->setStyle($rowStyleItalicBold);
    ->build()
```

niclaslindberg:
Hi @madflow and @adrilo  
Thanks for the good work. Just wanting to know if there is a plan to release @madflow's work of the ability to style separate cells. :)

adrilo:
Hi @niclaslindberg, the plan is definitely to release this. It will be released as part of v3 (which includes some non backwards compatible changes). This change will get merged as soon as we have time to get back to it. Other changes will follow to match this new way of using Spout (for the writer mainly).

Once these changes are committed, we'll release v3 as the new official version.

niclaslindberg:
Hi @adrilo 
Thanks for your answer. Sounds great! Is there a somewhat stable beta-branch for v3 that we can develop against? Thanks again.

adrilo:
@niclaslindberg `develop_3.0` is the official dev branch for v3

adrilo:
You're not missing anything. This PR is the first one related to the topic. Being able to style cell will only be possible once this commit gets merged.

davidNieves:
Hi!

I downloaded the last version of Spout and tried do this example, I get this error: 
"Fatal error: Class 'Box\Spout\Writer\Common\Entity\Row' not found "

The folder "Entity" is not in zip, what I have to do??

Thanks!!

madflow:
@davidNieves The changes currently only exists in my fork in the "cell_styling" branch. https://github.com/madflow/spout/tree/cell_styling

I would argue that a test would be done prematuraly - except you would like to help implement it. Then I would accept PRs to this branch in my fork.

src/Spout/Writer/ODS/Manager/WorksheetManager.php
madflow: @TODO - I marked this as a future refactoring step

src/Spout/Writer/WriterAbstract.php
madflow: But then the seconds argument  ```$baseStyle``` would need to be set to default ```null```. Wouldn't it?

src/Spout/Writer/XLSX/Manager/WorksheetManager.php
madflow: Also marked as future refactoring

madflow:
@adrilo 

I have been working on this PR again before it gets stale. Sorry for the delay (all types of excuses go here, Work-Life-Balance Fugazi etc.)


In the first step I deleted the *withStyle methods and made the tests pass again. However I have been scratching my head over the question, how the public ```Writer``` Api will look in the end with the newly added ```CreatorFactory->Entity<->Manager``` concept.

I looked at the ```Sheet``` entity and implemented the ```Row``` entity similiar.

When I would go down this path, we will have something like this in the documentation:

```php
$writer = WriterFactory::create(Type::XLSX);
$writer->setDefaultRowStyle($defaultStyleBorderBottom);
$writer->openToFile(__DIR__ . '/test.xlsx');

$entityFactory = new EntityFactory(new ManagerFactory());

$row = $entityFactory->createRow([], null); // Or inject Cells[], Style directly
$row
 ->addCell(
     $entityFactory->createCell('String Value')
 );

$writer->addRow($row);

$writer->close();
```

To me understanding - the "managers" are injected in the "entities" via factory methods. If you do not want to use the factory - it gets cumbersome.

Is this would you had in mind? I need some guidance here ;)

adrilo:
Don't worry about the delay. We're all pretty busy with other things :)

The dependency injection should be not be exposed externally. Unless you want developers to inject a specific behavior into the library, there's no need to expose this mechanism.
You can take a look at what I did in `WriterFactory`. The `WriterFactory` is what external developers interact with. It abstracts away the creation of all internal objects, for example:
```
    /**
     * @return XLSX\Writer
     */
    private static function getXLSXWriter()
    {
        $styleBuilder = new StyleBuilder();
        $optionsManager = new XLSX\Manager\OptionsManager($styleBuilder);
        $styleMerger = new StyleMerger();
        $globalFunctionsHelper = new GlobalFunctionsHelper();

        $entityFactory = new EntityFactory(new ManagerFactory());
        $internalFactory = new XLSX\Creator\InternalFactory($entityFactory);

        return new XLSX\Writer($optionsManager, $styleMerger, $globalFunctionsHelper, $internalFactory);
    }
```

So in your case, you could have something like `RowFactory::create()` or `EntityFactory::createRow`. This function is then responsible for creating a `ManagerFactory` object and inject it in other objects that will need it.

Hopefully that helps. Let me know if you still have some questions :)
Also what we can do is : we can merge your code even though it does not perfectly fit the new structure, and I'll refactor your code later. I don't want this to be a blocker for you, especially since we're on a develop branch!

madflow:
So the code I posted above would be the example for a possible documentation how developers would use the writer, correct? 

I already implemented ```EntityFactory::createRow``` and amended the tests to use it. What is missing is the ```EntityFactory::createCell``` refactoring. 

And - Yes we should merge better sooner than later. For instance I "hacked" the tests to pass and the de-hacking should be another step. There are some more questions that will arise... ;)

adrilo:
Actually, EntityFactory::createRow should be a static method so that we control the dependency injection. So your example would be more like this:

```
$writer = WriterFactory::create(Type::XLSX);
$writer->openToFile(__DIR__ . '/test.xlsx');

$row1 = EntityFactory::createRow();
$row1->addCell(EntityFactory::createCell('String Value'));
$writer->addRow($row1);

$row2 = EntityFactory::createRow([EntityFactory::createCell('String Value 1'), EntityFactory::createCell('String Value 2')]);
$row1->addCell(EntityFactory::createCell('String Value', $style));
$writer->addRow($row2);

$row3 = EntityFactory::createRow([EntityFactory::createCell('String Value', $style));
$writer->addRow($row3);

$writer->close();
```

tlaverdure:
Any idea on when cell styling will be ready? It's the final piece that I need to complete a project :)

Thanks

adrilo:
@tlaverdure no estimate so far. This commit may be merged soon but the v3 it will be part of won't be released any time soon, as there's still a lot of things to do.
However, feel free to use this commit as soon as it's merged. Although it won't be the official version, Spout should still be usable

madflow:
I merged the latest develop_3 branch and resolved the conflicts. I made the Row/Cell creators static. The old tests pass - I have not written specific new ones for the cell styling.

Travis failed because of the CS rules so I ran php-cs-fixer.

If this is the way to go - this PR should be merged so we can continue with smaller followup tasks.

src/Spout/Common/Helper/FileSystemHelper.php
```diff
@@ -61,7 +61,7 @@ public function createFileWithContents($parentFolderPath, $fileName, $fileConten
         $filePath = $parentFolderPath . '/' . $fileName;
 
         $wasCreationSuccessful = file_put_contents($filePath, $fileContents);
-        if ($wasCreationSuccessful === false) {
+        if (false === $wasCreationSuccessful) {
```
adrilo: Is this a php-cs-fixer rule? I don't really like this order... I think it's harder to read this way

madflow: Yes. I believe it is called "Yoda Style": https://en.wikipedia.org/wiki/Yoda_conditions.  See here: https://github.com/FriendsOfPHP/PHP-CS-Fixer

Maybe you can adjust the style settings to your liking? I personally always write boolean and === null expression this way. But I do not care, really...

adrilo: Oh I see. You actually need to pass it the actual config. I just updated it to disable yoda style : #472

adrilo: Also, you can run the fixer this way: `./vendor/bin/php-cs-fixer fix --config=.php_cs.dist`

madflow:
I updated the php-cs-fixer commit with the latest changes in the settings definition.

tlaverdure:
@adrilo ok cool, can't wait to try it out.

adrilo:
Hey @madflow, I'm planning to fix few minor things in this PR before merging. And agree, from there we'll be able to improve what has been done so far.
If it's not already done, do you mind giving me access to modify your PR (info here: https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)

Thanks!

tests/Spout/Writer/ODS/WriterWithStyleTest.php
```diff
@@ -56,42 +66,59 @@ public function dataProviderForInvalidStyle()
...
 
     /**
      * @dataProvider dataProviderForInvalidStyle
-     * @expectedException \Box\Spout\Common\Exception\InvalidArgumentException
      *
      * @param \Box\Spout\Writer\Common\Entity\Style\Style $style
      */
     public function testAddRowWithStyleShouldThrowExceptionIfInvalidStyleGiven($style)
     {
+        if (version_compare(PHP_VERSION, '7.0.0') >= 0) {
```
adrilo: why did you have to add that?

madflow: Because PHP < 7 does not know what a ```TypeError``` is.  This is one of the things I would like to adress in the future. "How should we handle Type Errors across PHP Versions etc".

madflow:
I have activated "Allow changes from maintainer".

tests/Spout/Writer/ODS/WriterWithStyleTest.php
adrilo: I guess the best way to handle that would be to only support PHP >= 7 :)

madflow: We have migrated our product to PHP 7.1.. So I would be full in :D

tests/Spout/Writer/XLSX/WriterTest.php
```diff
@@ -320,6 +344,7 @@ public function testAddRowShouldSupportAssociativeArrays()
      */
     public function testAddRowShouldNotWriteEmptyRows()
     {
+        $this->markTestIncomplete('Unsure why this does not pass');
```
adrilo: Actually, this is caused by a condition change is AbstractWriter. `!empty($dataRow)` is different from `$row->isEmpty()`. The former returns false for `['']` while the latter returns true.

adrilo:
Hey @madflow, I've done some changes but I need to rebase to merge. I tried multiple times but that's a nightmare... Unless you want to give a try, I think I'm going to import all your changes in a new PR. That's probably going to be easier!

madflow:
Have you had any luck with the changes in a new PR ?

adrilo:
I haven't tried doing it in a new PR yet. But I can definitely give it a try.
The problem is that when rebasing, it applies commit by commit and a lot of things changed between the first commits and now...

lflier:
Hi, is this implemented yet? I really need the individual cell styling and will likely add more options, too.

adrilo:
@lflier Not yet

lflier:
So it’s not working properly in this branch?

adrilo:
It should work, but be careful with it. The branch (and its target branch, v3.0) are still under development so things can change and eventually break your code.

adrilo:
@madflow I just created a new PR with all your changes, but on an up-to-date branch :) See #485 
Closing this PR then.

madflow:
:+1:

PR #439, opened 2017-06-01
HHVM is no longer supported on Ubuntu Precise

madflow:
Travis says:

> HHVM is no longer supported on Ubuntu Precise. Please consider using Trusty with `dist: trusty`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,13 @@
 language: php
 
+dist: trusty
+
 php:
   - 5.4
   - 5.5
   - 5.6
   - 7.0
-  - hhvm
+  - hhvm-3.6
 
 cache:
   directories:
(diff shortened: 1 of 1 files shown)


boxcla:

madflow:
I had to pin HHVM to 3.6 (the latest version that would run on Ubuntu Precise before). "hhvm" on "Trusty" " translates to "3.18" on Travis - which failed with:

> Fatal error: Syntax only allowed in Hack files (<?hh) or with -v Eval.EnableHipHopSyntax=true in /home/travis/build/box/spout/vendor/phpunit/phpunit-mock-objects/src/Framework/MockObject/Generator.php(290)(fff4f691f5d55d875f4d67761cc16122) : eval()'d code on line 539

Mehh...

PR #443, opened 2017-06-26
Allow setting of header content type for CSVs.

aronbudinszky:
Setting a different header content type can be useful (for CSVs) if different encoding is used for example.

### DIFF
--- a/src/Spout/Writer/AbstractWriter.php
+++ b/src/Spout/Writer/AbstractWriter.php
@@ -37,7 +37,7 @@ abstract class AbstractWriter implements WriterInterface
     protected $defaultRowStyle;
 
     /** @var string Content-Type value for the header - to be defined by child class */
-    protected static $headerContentType;
+    protected $headerContentType;
 
     /**
      * Opens the streamer and makes it ready to accept data.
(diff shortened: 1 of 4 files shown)


boxcla:
Hi @aronbudinszky, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

aronbudinszky:
Did not file an issue about this, but was really short to implement, so no crying if it is rejected. :)

aronbudinszky:
CLA signed

boxcla:
Verified that @aronbudinszky has just signed the CLA. Thanks, and we look forward to your contribution.

adrilo:
Hi @aronbudinszky, thanks for the PR!

By default, the writer generates CSV files encoded in UTF-8, with a BOM. It is possible to remove the BOM, but the file will still be UTF-8 encoded. So there's no need to override the charset.

And even if you would want to do it, I would instead override it manually:
```
$writer->openToBrowser('...');
header('Content-Type: 'text/csv; charset=UTF16-LE', true); // the "true" will override previously set headers
```

PR #457, opened 2017-08-26
Refactor readers for better di

adrilo:
Refactor readers to get a proper DI:
Similar to what was done with writers, readers also needed to be updated to match the new way of doing things. This commits promotes a better DI (factories, injection through constructors).

Escapers should not be singletons:
Instead, they should be proper object that can be injected where needed.


boxcla:

PR #459, opened 2017-08-27
Favor object creation in factories

adrilo:
Instead of passing factories in the constructors and let objects call the factory method, create all dependencies directly in the factories.


boxcla:

PR #460, opened 2017-08-27
Introduce Managers for readers

adrilo:
Some helper classes were more managers than helpers...


boxcla:

PR #475, opened 2017-10-11
As per RFC-4180 regarding CSV, "If double-quotes are used to enclose …

garimpeirodouniverso:
…fields, then a double-quote appearing inside a field must be escaped by preceding it with another double quote".

So we are changing the calls to fgetcsv and fputcsv on GlobalFunctionsHelper in order to send the Escape Character as a parameter. To be RFC compliant, we always send the same character as the Field Enclosure character.

### DIFF
--- a/src/Spout/Common/Helper/GlobalFunctionsHelper.php
+++ b/src/Spout/Common/Helper/GlobalFunctionsHelper.php
@@ -86,7 +86,7 @@ class GlobalFunctionsHelper
      */
     public function fgetcsv($handle, $length = null, $delimiter = null, $enclosure = null)
     {
-        return fgetcsv($handle, $length, $delimiter, $enclosure);
+        return fgetcsv($handle, $length, $delimiter, $enclosure, $enclosure);
     }
 
     /**
@@ -101,7 +101,7 @@ class GlobalFunctionsHelper
      */
     public function fputcsv($handle, array $fields, $delimiter = null, $enclosure = null)
     {
-        return fputcsv($handle, $fields, $delimiter, $enclosure);
+        return fputcsv($handle, $fields, $delimiter, $enclosure, $enclosure);
     }
 
     /**


boxcla:
Hi @garimpeirodouniverso, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

garimpeirodouniverso:
CLA signed!

boxcla:
Verified that @garimpeirodouniverso has just signed the CLA. Thanks, and we look forward to your contribution.

madflow:
https://github.com/facebook/hhvm/commit/85d3041838dce60e16987978c604105a0b731288#diff-0cb218f3ae8ba68e5cdb59a23a6ce0bb

Our HHVM version on travis is 3.6. I think we need at least 3.11 to make this work. I can take a look at the Travis build - i pinned it to that version. I forgot why I did it though...

madflow:
Now I remember what the deal with HHVM was: 3.6 was the one of the last versions the unit tests would pass. After updating to a version that supports the excape char HHVM yields:

```Fatal error: Syntax only allowed in Hack files (<?hh) or with -v Eval.EnableHipHopSyntax=true in /home/travis/build/madflow/spout/vendor/phpunit/phpunit-mock-objects/src/Framework/MockObject/Generator.php(345)(eb39396dd37ca5560e6fe5505b82153b) : eval()'d code on line 540```

This is due to https://github.com/sebastianbergmann/phpunit-mock-objects/blob/3.4/src/Framework/MockObject/Generator.php#L345

Before talking about possible hacks and fixes - since HHVM is not targeting PHP features in the future (http://hhvm.com/blog/2017/09/18/the-future-of-hhvm.html) dropping HHVM from the test grid should be considered @adrilo .

**Comment by adrilo** [comment]
I'm fine with dropping HHVM support. It will be too painful to maintain both systems.

**Comment by adrilo** [comment]
@garimpeirodouniverso can you please add a test case that covers this change?

**Comment by garimpeirodouniverso** [comment]
@adrilo I can for sure, but I'm afraid you'll have to guide me on how to accomplish this... Can you kindly?

**Comment by adrilo** [comment]
Sure! You can create a CSV test file with the values you want to test (so probably a value "escaped with `\` and one escaped with the enclosure). You can add it in the `tests/resources/csv` folder, using a proper name.
Then, add a test in this file: https://github.com/box/spout/blob/master/tests/Spout/Reader/CSV/ReaderTest.php. You can look at other tests and just adapt them to do what you want.

Let me know if you need additional help

**Comment by adrilo** [comment]
Also, make sure to rebase, as @madflow just pushed the "fix" for HHVM

**Comment by bugalot** [comment]
Hello there!

I was encountering the same issue and also thought that using similar enclosure and escape characters would solve my problems. Unfortunately, the following experiment did not make it.

```php
<?php
/*
Expected, non enclosed/escaped data is

col1,col2,
val1\,val2
"hello",world

*/

$data = [
    ["col1", "col2"],
    ["val1\\", "val2"],
    ["\"hello\"", "world"]
];

$delimiter = ',';
$enclosure = '"';
$escape = $enclosure;

$handle = fopen('data.csv', 'w');
foreach ($data as $d) {
    fputcsv($handle, $d, $delimiter, $enclosure, $escape);
}
fclose($handle);

$handle = fopen('data.csv', 'r');
$readData = [];
while($data = fgetcsv($handle, 0, $delimiter, $enclosure, $escape)) {
    $readData[] = $data;
}
fclose($handle);


/*
array(3) {
  [0]=>
  array(2) {
    [0]=>
    string(4) "col1"
    [1]=>
    string(4) "col2"
  }
  [1]=>
  array(2) {
    [0]=>
    string(5) "val1\"
    [1]=>
    string(4) "val2"
  }
  [2]=>
  array(2) {
    [0]=>
    string(7) "hello"""
    [1]=>
    string(5) "world"
  }
}
*/
var_dump($readData);
```

As you can see, after reading the data, the expected value `"hello"` is read as `hello""`. In such a context, it seems to me that using a similar value for the enclosure and escape characters does not solve everything.

I hope I missed something :smile: !

adrilo:
@bugalot thanks for you input!
Just to make sure we are on the same page here, what should the CSV file look like? Should it be this?
```
col1,col2
val1\,val2
""hello"",world
```

adrilo:
@bugalot actually there was something wrong with the way your CSV file was generated.
If you want to get `"hello"` then your CSV file should contain `"""hello"""` (3 quotes) and not `""hello""` (2 quotes).
We can therefore move forward with this change :)

adrilo:
I took @garimpeirodouniverso's changes and will merge them as part of #500

bugalot:
You are right @adrilo !

However, if my raw CSV input file looks like:

```
col1,col2
val1\,val2
"""hello""",world
```

The final var_dump of the example above now returns

```
array(3) {
  [0]=>
  array(2) {
    [0]=>
    string(4) "col1"
    [1]=>
    string(4) "col2"
  }
  [1]=>
  array(2) {
    [0]=>
    string(5) "val1\"
    [1]=>
    string(4) "val2"
  }
  [2]=>
  array(2) {
    [0]=>
    string(7) "hello"""
    [1]=>
    string(5) "world"
  }
}
```

So it's seems I receive the `hello""` string rather than `"hello"` :cry:

PR #476, opened 2017-10-13
(chore) remove hhvm from the test grid

madflow:


boxcla:

PR #483, opened 2017-11-04
Add support for 1904 dates

### LINKED ISSUE #480: 1904-based dates are parsed incorrectly
1904-based dates are common on Mac. The offset from the 1900-based dates is 1462 days.

[What they are and why](https://support.microsoft.com/en-us/help/214330/differences-between-the-1900-and-the-1904-date-system-in-excel).

[How to tell which format is used (on PHPExcel's example)](https://github.com/ddeboer/phpexcel/blob/master/Classes/PHPExcel/Reader/Excel2007.php#L716)

adrilo:
Fixes #480 

This commit adds support for dates using the 1904 calendar (starting 1904-01-01 00:00:00).
It also fixes some issues with the dates in 1900 calendar (which now correctly start at 1899-12-30 00:00:00).
Finally, it is now possible to have negative timestamps, representing dates before the base date (and up to 0000-01-01 00:00:00), as per the SpreadsheetML specs. Note that some versions of Excel don't support negative dates...


boxcla:

Gavrisimo:
Hey @adrilo, is there any reason why this can't be added to v2 as well?

adrilo:
The only reason is a lack of time :) The underlying implementation changed quite a bit between v2 and v3 so I focused on the v3 part.

Gavrisimo:
Ok, I think we'll just make a switch to v3 and see how things are working out.

PR #485, opened 2017-11-04
Row objects and Cell styling - v2

adrilo:
This commit introduces Row and Cell entities, that will replace the arrays passed in previously.
It also adds support for Cell styling (instead of Row styling only).


boxcla:

PR #490, opened 2017-11-05
Consolidate external EntityFactory

adrilo:
All entities will now be created through a single factory (including the Writers).
Also, added a EntityFactory::createRowFromArray() to make it easier to create rows


boxcla:

Issue #492, opened 2017-11-08
(develop) Always write rows


madflow:
Currently there are at least two checks if the provided row should be added to the writer. For example in the XLSX writer:

https://github.com/box/spout/blob/8aec9ea9929965b483361e89db67a62828c9f2f4/src/Spout/Writer/WriterAbstract.php#L182

and 

https://github.com/box/spout/blob/8aec9ea9929965b483361e89db67a62828c9f2f4/src/Spout/Writer/XLSX/Manager/WorksheetManager.php#L124

I would vote for the ```develop``` branch to drop theses concepts altogether. Skipping data should be the responsibilty of the developer.

An empty row can be desirable. So a developer could add a ```Row``` with not cells and this would create an empty row in the Spreadsheet or add only empty cells.

adrilo:
Hey @madflow !
I'm fine with that proposal. Although only the first snipped should be removed. The second one is here so that we don't write any row in the XML file if there are no data (just an size/speed optimization).

madflow:
I would personally remove both. In my opinion the writer library should not decide what an empty row is. This is the current backported check:

```PHP
    /**
     * Detect whether a row is considered empty.
     * An empty row has either no cells at all - or only one empty cell
     *
     * @param Row $row
     * @return bool
     */
    public function isEmpty(Row $row)
    {
        $cells = $row->getCells();
        return count($cells) === 0 || (count($cells) === 1 && $cells[0]->isEmpty());
    }
```

So if I want to write an empty row - I would have to create two empty cells... even if I have just one column. 

In addition: I cannot really see the performance penalty here. Maybe I am missing something ...

adrilo:
Actually, keeping this check would not prevent you from adding an empty row. The optimization I was talking about is that if you have an empty row (a row with no cell let's say), then there's no point writing anything to the XML file. The only thing you need to do is increment the row number so that the next row you'll write will be after this blank row. And this is already what's being done in the `addRow()` function you mentioned.
So to me, this check is totally valid and will reduce the final size of the XML file as well as the time it takes to generate it.

madflow:
Thanks for the clarification. The second check can stay :)

PR #493, opened 2017-11-10
Always write rows

madflow:
+ always write rows even with no cells #492 
+ add upgrade documentation #492


boxcla:

adrilo:
👍

PR #495, opened 2017-11-10
StyleMerger and RowManager changes

adrilo:
Move DI of StyleMerger to get rid of business logic in the entities Cell and Row
Simplify RowManager


boxcla:

PR #499, opened 2017-11-11
Fetch XML file paths from Workbook Relationships

### LINKED ISSUE #441: Crashes while reading table cells from some xslx-files
Some xslx archives have the shared strings file which name starts with an uppercase letter (SharedStrings.xml). In this case Spout crashes with an error while reading a table cell: 

`Fatal error: Call to a member function getStringAtIndex() on null in Spout/Reader/XLSX/Helper/SharedStringsHelper.php on line 249`

adrilo:
Fixes #441 

The file path of the XML files inside the XLSX files were hardcoded. However, in some cases the file name did not match the hardcoded one (different case for instance). So instead of relying on a hardcoded path, read it from the `workbook.xml.rels` file.


boxcla:

PR #500, opened 2017-11-11
Fix charachters escaping with CSV reader/writer

### LINKED ISSUE #473: Backslash on CSV is escaping the double quote but it should not
I have something like the following CSV file:

"col1";"col2"
"val1\\";"val2"

The expected values from the second row would be _val1\\_ and _val2_ , because the backslash should not escape the double quote (the correct way of escaping the double quote would be to "double-double quote" instead, as per RFC-4180).

Instead, Spout will read it as _val1\\";val2"_ (as the first value), and ignore the second value.

Am I missing something? Is there a way to set the backslash as a non-escaper?

Thanks

adrilo:
Fixes #473 

PHP's built-in functions fputcsv and fgetcsv are not RFC-4180 compliant and include an escape character that's not defined in the spec. This results in escaping characters that should not be escaped.
This commit disables this escaping mechanism.


boxcla:

garimpeirodouniverso:
Hello @adrilo , sorry for not being able to help recently on the tests for the fix. Good news you guys are incorporating this on 3.0!

Are there any estimates for this release?

adrilo:
No ETA yet. This is what is remaining: https://github.com/box/spout/milestone/1
We may add a few simple features on top of that but for now, the focus should be on #501 as that's the biggest one left. Since we all maintain this library on our free time, it's very hard to say when we'll be done with 3.0.

Issue #463, opened 2017-08-27
[Docs] Move documentation from gh-pages to "docs" folder


adrilo:
When documentation is on the gh-pages branch, it's not possible to version it.
Since it will change a lot with Spout v3, it would be good to keep the documentation with the code. It is also true for minor changes!

madflow:
You assigned it to yourself - but I can take a look at it, too... I would have to know it there still should be a Jekyll compatible deployment for gh-pages with the "docs" folder as source. Your call.

adrilo:
Easiest thing would be to just move what's in the `gh-pages` branch into the `docs` folder. I think we can do that for now.
The main problem here is that I no longer have admin access to the repo so I can't freely change this setting... I guess this can be done in a 2nd step.

PR #516, opened 2017-11-21
#463, move gh-pages to docs folder

madflow:


boxcla:

PR #523, opened 2017-12-15
Clean else

carusogabriel:
I've removed the `else` when we have already returned something :put_litter_in_its_place:


boxcla:
Verified that @carusogabriel has signed the CLA. Thanks for the pull request!

PR #535, opened 2018-02-08
Row::getCell

madflow:


boxcla:

src/Spout/Common/Entity/Row.php
```diff
@@ -65,6 +64,15 @@ public function setCellAtIndex(Cell $cell, $cellIndex)
         return $this;
     }
 
+    /**
+     * @param int $cellIndex
+     * @return Cell|null
+     */
+    public function getCell($cellIndex)
```
adrilo: for consistency, I would rename to `getCellAtIndex`

madflow:
Updated the method name and rebased

PR #543, opened 2018-03-23
fix risky tests and assert true for silent tests

madflow:


boxcla:

adrilo:
👍

PR #559, opened 2018-06-03
Upgrade to 3.0 guide

### LINKED ISSUE #503: Update migration guide v2 => v3

adrilo:
Fixes #503


adrilo:
@madflow could you take a quick look and let me know if you think of another breaking change I'm missing.

madflow:
Looks good to me. As far as I can see the breaking changes are mostly API changes so far. When devs migrate to the Creator/Writer/Reader/Book/Sheet/Rows/Cells model -  they should get the same result. (Or already stated otherwise in this doc). :+1:

PR #565, opened 2018-06-12
Remove HHVM workarounds

madflow:
* HHVM is removed from the test grid and if we want to use PHP 7.1 features - it will have to go for good.

### DIFF
--- a/src/Spout/Common/Helper/Escaper/ODS.php
+++ b/src/Spout/Common/Helper/Escaper/ODS.php
@@ -9,40 +9,22 @@ namespace Box\Spout\Common\Helper\Escaper;
 class ODS implements EscaperInterface
 {
     /**
-     * Escapes the given string to make it compatible with XLSX
+     * Escapes the given string to make it compatible with ODS
      *
      * @param string $string The string to escape
      * @return string The escaped string
      */
     public function escape($string)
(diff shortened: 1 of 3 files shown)


boxcla:

madflow:
Okay - I see there is more to it:  https://github.com/box/spout/blob/b105d15f08ccb62240945096099447c8963d63a1/src/Spout/Reader/Wrapper/XMLReader.php#L23

madflow:
@adrilo Is there anymore work to do? I looked into the XMLReader due to to the comment ``` The associated code in XMLReader::open()... ```. As far as I can tell the code in XMLReader has to stay the same and cannot be removed.

adrilo:
Hey @madflow !

I think you covered all the places that needed to be updated to remove support for HHVM.
But before we go ahead, I'm still wondering if that's a good idea to totally remove support for it. I don't think the features we use in PHP 7.1 are incompatible with HHVM and keeping what we added specifically for HHVM does not really hurt. Worst case, we break the lib for HHVM but we can claim we have no official support for it. And we can let the community provide fixes for it if needed.
So I'd be in favor of doing nothing here. What do you think?

madflow:
Sounds good to me!

Issue #561, opened 2018-06-08
Reader doesn't understand which is the last column (ODS file)


eugenio11:
Hello,
for some files (until now, it happened only with ODS files), it seems the reader wrongly see all the 1024 columns as part of the sheet, even if just three columns are used. Basically the reader, from the fourth column on, returns an empty string for every cell.

I am not sure if it's a spout issue and most of the time it doesn't happen, I cannot reproduce the issue starting from a new ODS file but I have an ODS file that triggers the issue, if you need it. 

Thanks.

madflow:
@eugenio11 Yes - please provide the file in an anonymized form if possible

eugenio11:
[tet.ods.zip](https://github.com/box/spout/files/2086797/tet.ods.zip)

Hello, here is the file

adrilo:
Hi @eugenio11 !

Here is what the content file of the ODS looks like:
```
<table:table-row table:style-name="ro1">
	<table:table-cell office:value-type="string">
		<text:p>nome</text:p>
	</table:table-cell>
	<table:table-cell office:value-type="string">
		<text:p>cognome</text:p>
	</table:table-cell>
	<table:table-cell office:value-type="string" table:style-name="ce3">
		<text:p>indirizzo</text:p>
	</table:table-cell>
	<table:table-cell table:number-columns-repeated="1021"/>
</table:table-row>
```

The last `table-cell` line is what causes the 1000+ columns to be added. The fix should be to ignore the line if it's the last one in the row, and if the value is empty.

eugenio11:
Hello and thanks for the reply.
I am wondering how a file like the one I attached can be produced, I don't remember the steps I did to make it and I am not able to reproduce it.
About your fix, what if all the 1000+ columns are actually used?

adrilo:
Some softwares do add this repeated columns cell for some reasons. Some versions of Excel for instance add this. I'm not sure why...
I made the change but ended up breaking some tests, which reminded me that your point is good. Sometimes, it is the desired behavior to have empty cells at the end of the row.

I'm gonna look at the official specs to see what we should do in this case. Eventually, we could have something a bit magic: if the last cell is empty and repeated more than X times, then do not add it; X being a magic number to be determined. This approach should solve 99% of the use cases and be the cause of bugs in the 1% of cases where the last cell is actually repeated a lot on purpose.

madflow:
[...]

PR #568, opened 2018-06-15
Introduce a Reader range

madflow:
This PR introduces  a range for readers like discussed here: https://github.com/box/spout/issues/561#issuecomment-396159208

The range can be defined on a ```Reader``` by setting a zero based start - and end column index. This can be done before or after a reader has been opened.

The  reader will only yield data starting at the start index - and stop reading **after** the end index.

If an end index is provided missing cells are filled with empty values so there is always the same range.

**Given the CSV:**

```CSV
'csv--11', 'csv--12', 'csv--13'
'csv--21', 'csv--22'
'csv--31'
```

```PHP
$reader->open($csv);
$reader->setStartColumnIndex(0);
$reader->setEndColumnIndex(1);
```

**Outputs**:
```
array(3) {
  [0]=>
  array(2) {
    [0]=>
    string(7) "csv--11"
    [1]=>
    string(7) "csv--12"
  }
  [1]=>
  array(2) {
    [0]=>
    string(7) "csv--21"
    [1]=>
    string(7) "csv--22"
  }
  [2]=>
  array(2) {
    [0]=>
    string(7) "csv--31"
    [1]=>
    string(0) ""
  }
}

```

I only did this for the CSV reader for ```RFC``` purposes.

### DIFF
--- a/src/Spout/Common/Manager/OptionsManagerAbstract.php
+++ b/src/Spout/Common/Manager/OptionsManagerAbstract.php
@@ -7,7 +7,7 @@ namespace Box\Spout\Common\Manager;
  */
 abstract class OptionsManagerAbstract implements OptionsManagerInterface
 {
-    const PREFIX_OPTION = 'OPTION_';
+    public const PREFIX_OPTION = 'OPTION_';
 
     /** @var string[] List of all supported option names */
     private $supportedOptions = [];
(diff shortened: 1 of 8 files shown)


boxcla:

madflow:
Maybe naming the exposed setter "setRange" and keeping the columnindexes internal would be a better option...?

madflow:
I have no plans on working on this PR anymore. Closing...

Issue #526, opened 2017-12-21
[DEV] Entities, Writer and Reader factories


madflow:
In the ```develop_3.0``` branch the writer and reader instances can be created through 

* `Box\Spout\Reader\Common\Creator\EntityFactory`
* `Box\Spout\Writer\Common\Creator\EntityFactory`

in addition there is a ```WriterFactory``` but not a ```ReaderFactory``` in the same namespaces. 

I was taking a look at the docs and the first thing I did was this:

```
use Box\Spout\Reader\Common\Creator\EntityFactory as ReaderFactory;
use Box\Spout\Writer\Common\Creator\EntityFactory as WriterFactory;
```

So I immediatly tried to alter the meaning without thinking about it ;)  Before the docs are brought up to speed with the latest development I would like to disuss this shortly.

The `Reader` and the `Writer` do not look like entities to me (or more precise what I think an entity is). An entity to me is something you can talk about like: A spreadsheet has sheets. A sheet contains columns, rows and cell etc. Someone who kows Excel will know what you are talking about, even without knowledge of spout. It is hard to talk about a business logic monster like ```Writer\XLSX```  in that way.

I would argue that the entitiy factory should only create "entities" and the Writer and Reader should be separate. This would also be an easier migration path. What do you think?

adrilo:
Finally coming back to this issue.

> in addition there is a WriterFactory but not a ReaderFactory in the same namespaces.

Good catch, I will fix that (I guess the Reader section needs a bit of work to match the "Writer model")

Regarding the use of `EntityFactory` and `Reader/WriterFactory`, I think it over complicates things to have to deal with multiple factories to create the different kind of objects that could be needed.
Such a code would look like:
```
$writer = WriterFactory::create('xlsx');
$writer->addRow(EntityFactory::createRow([
    EntityFactory::createCell(...),
    EntityFactory::createCell(...),
]));
```
I don't believe this is very practical. Also, the migration from v2 to v3 needs many changes. Not changing the name of the factory does not help that much, as the namespace did change, so devs will need to adapt to that.

I do like having a central "interface" which handles the creation of all the objects, even though I agree that Readers and Writers are not really "entities". I picked this term because I was lacking of a better one. If you have any idea, I'm all ears!

adrilo:
I just realized the issue with the EntityFactory naming. Having 2 classes with exactly the same name forces devs to do what you did (i.e. using an alias) if they need both the reader and the writer.
[...]

PR #573, opened 2018-06-28
rename EntityFactory for writers and readers #526

madflow:
#526


boxcla:

PR #577, opened 2018-07-03
Fix sheet name escaping (fixes #576)

### LINKED ISSUE #576: Incorrect escaping used for Excel and ODS spreadsheet names
In Excel, "test" is a valid spreadsheet name, and this needs to be saved in workbook.xml as

    <sheets><sheet name="&quot;test&quot;" sheetId="1" r:id="rId1"/></sheets>

However, this library gets that wrong. That is, this comment: https://github.com/box/spout/blob/master/src/Spout/Common/Escaper/XLSX.php#L45 and the following code is incorrect.

Ditto for ODS.

I am about to make a pull request for this.

timhunt:
I hope this is OK. This is the first time I have contributed to this project. Please let me know if I need to change anything. Thanks in advance for considering this.

### DIFF
--- a/src/Spout/Common/Escaper/ODS.php
+++ b/src/Spout/Common/Escaper/ODS.php
@@ -22,17 +22,17 @@ class ODS implements EscaperInterface
      */
     public function escape($string)
     {
+        // @NOTE: ODS files are basically XML, so we need to escape <, > and &.
+        // Mostly values are saved in element contents, but a few things (e.g. worksheet names)
+        // are stored in attributes, so we also need to escape " and ', hence ENT_QUOTES.
         if (defined('ENT_DISALLOWED')) {
             // 'ENT_DISALLOWED' ensures that invalid characters in the given document type are replaced.
             // Otherwise control characters like a vertical tab "\v" will make the XML document unreadable by the XML processor
             // @link https://github.com/box/spout/issues/329
-            $replacedString = htmlspecialchars($string, ENT_NOQUOTES | ENT_DISALLOWED, 'UTF-8');
+            $replacedString = htmlspecialchars($string, ENT_QUOTES | ENT_DISALLOWED, 'UTF-8');
         } else {
             // We are on hhvm or any other engine that does not support ENT_DISALLOWED.
-            //
-            // @NOTE: Using ENT_NOQUOTES as only XML entities ('<', '>', '&') need to be encoded.
-            //        Single and double quotes can be left as is.
-            $escapedString =  htmlspecialchars($string, ENT_NOQUOTES, 'UTF-8');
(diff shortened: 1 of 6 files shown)


boxcla:
Verified that @timhunt has signed the CLA. Thanks for the pull request!

src/Spout/Common/Escaper/XLSX.php
```diff
@@ -42,9 +42,9 @@ protected function init()
     public function escape($string)
     {
         $escapedString = $this->escapeControlCharacters($string);
-        // @NOTE: Using ENT_NOQUOTES as only XML entities ('<', '>', '&') need to be encoded.
-        //        Single and double quotes can be left as is.
-        $escapedString = htmlspecialchars($escapedString, ENT_NOQUOTES, 'UTF-8');
+        // @NOTE: We must use ENT_QUOTES because single and double quotes may be used in spreadsheet names.
```
adrilo: Can you please change that note to something more global like this: "Using ENT_QUOTES as XML entities ('<', '>', '&') as well as double quotes (for values of XML attributes) need to be encoded." ?

src/Spout/Common/Escaper/ODS.php
```diff
@@ -26,13 +26,13 @@ public function escape($string)
...
             // @link https://github.com/box/spout/issues/329
-            $replacedString = htmlspecialchars($string, ENT_NOQUOTES | ENT_DISALLOWED, 'UTF-8');
+            $replacedString = htmlspecialchars($string, ENT_QUOTES | ENT_DISALLOWED, 'UTF-8');
         } else {
             // We are on hhvm or any other engine that does not support ENT_DISALLOWED.
             //
-            // @NOTE: Using ENT_NOQUOTES as only XML entities ('<', '>', '&') need to be encoded.
-            //        Single and double quotes can be left as is.
-            $escapedString =  htmlspecialchars($string, ENT_NOQUOTES, 'UTF-8');
+            // @NOTE: We must use ENT_QUOTES because single and double quotes may be used in spreadsheet names.
```
adrilo: Can you please change that note to something more global like this: "Using ENT_QUOTES as XML entities ('<', '>', '&') as well as double quotes (for values of XML attributes) need to be encoded." ?

JasonEverling:
So can this bug get merged? It is affecting other software that use this library

adrilo:
Actually, looking at the test changes, this change is not ok: `'adam\'s "car"'` being transformed into `'adam&#039;s &quot;car&quot;'` is not something we want. We need to find another way to fix this issue. Maybe have different escape methods?

adrilo:
Closing this PR as I pushed another one here: #620 (was easier than resolving conflicts on your branch).

PR #584, opened 2018-09-03
trying to property of non-object is repaired

sasha-mart:
$tNode can be null if the cell is empty.


boxcla:
Verified that @sasha-mart has signed the CLA. Thanks for the pull request!

adrilo:
Do you have an example of such a file?

brosbunbo:
Why is this fix still not merged yet?

PR #593, opened 2018-10-08
use expectNotToPerformAssertions

madflow:
I introduced ```$this->assertTrue(true);``` hacks a while back to assert something so PHPUnit does not mark these tests as risky. I learned that there is a PHPUnit "way" to do this. by using ```$this->expectNotToPerformAssertions()```.


boxcla:

PR #594, opened 2018-10-15
adding xlsx support for column widths, merged cells, number formats

willkenson:


adrilo:
I don't think this was meant to be a PR but rather a bunch of commits in a fork. Closing

PR #604, opened 2018-12-01
Fix phpunit

yannickroger:
# Description
Phpunit would fail when I ran them locally. It was due to a dependency error.

So I enforced PHP 5.4 in `composer.json` (first commit) and ran a composer update.

Now it runs just fine.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,11 @@
         "branch-alias": {
             "dev-master": "2.8.x-dev"
         }
+    },
+    "config": {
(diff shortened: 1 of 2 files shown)


yannickroger:
CLA signed

boxcla:
Verified that @yannickroger has just signed the CLA. Thanks, and we look forward to your contribution.

adrilo:
👍 I'll do the same for the 3.0 branch

PR #605, opened 2018-12-01
 Allow getting an instance of a reader without Exception 

yannickroger:
# Description
When using the Reader factory you have to surround it with a try/catch since it can throw an exception.

If you hard code the Type in the call, there are no chance the exception is ever thrown, so you end up with a block of dead code for catching something that will never happen.

This PR adds dedicated method to retrieve an instance of a reader without having an exception thrown.

# Test
Updated test coverage

### DIFF
--- a/src/Spout/Reader/ReaderFactory.php
+++ b/src/Spout/Reader/ReaderFactory.php
@@ -29,20 +29,56 @@ class ReaderFactory
 
         switch ($readerType) {
             case Type::CSV:
-                $reader = new CSV\Reader();
+                $reader = self::createCsvReader();
                 break;
             case Type::XLSX:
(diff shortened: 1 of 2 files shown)


boxcla:
Verified that @yannickroger has signed the CLA. Thanks for the pull request!

yannickroger:
What do you think about that one @adrilo ?

adrilo:
I believe doing `createReader(Type:XLSX)` and `createXLSXReader()` is the same. You don't need to catch the exception if you know the type and hardcode it. Catching the exception can be useful if the type you pass to `createReader` is coming from a variable and may vary. But in that case, you won't be able to use your shortcuts neither.

PR #606, opened 2018-12-01
Current row not available w/o rewinding Iterator

yannickroger:
# Description
I ran into this issue while using the library. So here is a unit test illustrating it.

If you don't rewind the iterator after you got it (like a `foreach` would automatically), the current element is not set.

I might add the fix to this issue later, but I wanted you to have a look at this first as it seems the fix will involve quite a lot of changes on the internal of Reader.

# TODO
- [x] Unit test to highlight the issue
- [ ] Fix the issue

### DIFF
--- a/tests/Spout/Reader/CSV/ReaderTest.php
+++ b/tests/Spout/Reader/CSV/ReaderTest.php
@@ -515,4 +515,31 @@ class ReaderTest extends \PHPUnit_Framework_TestCase
         $reader->open('unsupported://foobar');
     }
 
+    /**
+     * @return void
+     */
+    public function testReadWithoutCallingRewindFirst()
+    {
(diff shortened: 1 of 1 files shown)


adrilo:
Thanks @yannickroger! There's indeed an issue and it is not limited to CSV (it applies to XLSX and ODS readers as well).
Fixing it would require quite some work so I'd suggest keeping this issue open and plan a fix for the next release (3.1). However, if you are willing to take a look, feel free to checkout the `develop_3.0` branch and make a PR in that branch to have the fix released with the 3.0 version.

adrilo:
Although it is valid, using `current()` on the iterator directly is not the usual/recommended way of using Spout. As there's an easy workaround (calling `rewind()` before `current()`), I don't think it's worth spending time fixing this issue.

PR #607, opened 2018-12-03
For PHP 7.1 for Composer

adrilo:
If local PHP is > 7.1, we still want to download dependencies compatible with PHP 7.1


boxcla:

adrilo:
Follow-up of #604

PR #612, opened 2019-01-03
Formula support

elibyy:
Hey there,
I always needed the formula in my reports, so I've made an implementation of that, And it would be nice to have it within the original package.

P.S. Really needed that on production ASAP so had to change the composer package, if this causes a problem with the PR, I can fork to different repo

Thanks
Eli

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "box/spout",
+    "name": "elibyy/spout",
     "description": "PHP Library to read and write spreadsheet files (CSV, XLSX and ODS), in a fast and scalable way",
     "type": "library",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "box/spout",
```
barryvdh: This shouldn't be changed in your PR

elibyy: Yes, as being said I needed this in prod, so I'll commit another with the correct package name

elibyy:
This ia based on that PR but with two changes
1. There's a class and not just guessing by equals sign
2. You can set the pre calculated value so if the user opens it in excel without enabling editing they would still have a value

barryvdh:
Should this be targeted at the v3 branch? As that is coming soon(ish) as it seems?

Slamdunk:
Merged a variation in https://github.com/openspout/openspout/pull/5

PR #620, opened 2019-01-26
Fix sheet name escaping

adrilo:
Fixes #576
Sheet names are stored as attributes of an XML entity. We therefore need a different escaping strategy, escaping quotes.

PR #625, opened 2019-02-02
Fix test failure message

adrilo:
Following up #620


boxcla:

PR #630, opened 2019-03-08
Add Factory Methods for each type of writers

sinri:
Add Factory Methods for each type of writers, to provide better experience with correct PHPDoc in IDE.

The correct factory provides writers of any type with PHPDoc of `WriterInterface`, it is not friendly for IDE to support the instance with its special methods. For example, IDE might warn you when you use `setShouldAddBOM` with a CSV writer if you generate it by the factory. Let IDE know what class you generated might be a kind of choice to refine.

### DIFF
--- a/src/Spout/Writer/WriterFactory.php
+++ b/src/Spout/Writer/WriterFactory.php
@@ -45,4 +45,34 @@ class WriterFactory
 
         return $writer;
     }
+
+    /**
+     * @return CSV\Writer
+     */
+    public static function createCSVWriter()
+    {
+        $writer = new CSV\Writer();
+        $writer->setGlobalFunctionsHelper(new GlobalFunctionsHelper());
+        return $writer;
+    }
+
+    /**
+     * @return XLSX\Writer
+     */
+    public static function createXLSXWriter()
+    {
+        $writer = new XLSX\Writer();
+        $writer->setGlobalFunctionsHelper(new GlobalFunctionsHelper());
+        return $writer;
+    }
+
+    /**
+     * @return ODS\Writer
+     */
+    public static function createODSWriter()
+    {
+        $writer = new ODS\Writer();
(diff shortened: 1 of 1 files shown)


boxcla:
Hi @sinri, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

sinri:
CLA signed

----
Note, The signing page returned me a white screen after my submit.

boxcla:
Verified that @sinri has just signed the CLA. Thanks, and we look forward to your contribution.

src/Spout/Writer/WriterFactory.php
```diff
@@ -45,4 +45,34 @@ public static function create($writerType)
 
         return $writer;
     }
+
+    /**
+     * @return CSV\Writer
+     */
+    public static function createCSVWriter()
```
adrilo: use these functions in the `create` method above.

adrilo:
Thanks @sinri! Can you make this PR against the develop_3.0 branch instead of master?

sinri:
I'll deal this later

src/Spout/Writer/WriterFactory.php
sinri: > method

Should I modify the `create` method?

If needed it could be done later.

sinri:
> Thanks @sinri! Can you make this PR against the develop_3.0 branch instead of master?

Emmm, I read new `WriterFactory` in the develop_3.0 branch, and I think the implementation could be more clearly dealt by make `get***Writer` methods public and make same static create methods in `WriterEntityFactory` class.

SEE https://github.com/box/spout/pull/633

src/Spout/Writer/WriterFactory.php
adrilo: Yes, please modify the `create` method to call the newly created methods directly.
It should be done now rather than later.

PR #633, opened 2019-03-24
According to https://github.com/box/spout/pull/630#issuecomment-47446…

sinri:
According to the requirement as

https://github.com/box/spout/pull/630#issuecomment-474463226

Originally PR to master now to dev 3 branch.

> Add Factory Methods for each type of writers, to provide better experience with correct PHPDoc in IDE.

> The correct factory provides writers of any type with PHPDoc of WriterInterface, it is not friendly for IDE to support the instance with its special methods. For example, IDE might warn you when you use setShouldAddBOM with a CSV writer if you generate it by the factory. Let IDE know what class you generated might be a kind of choice to refine.

### DIFF
--- a/src/Spout/Writer/Common/Creator/WriterEntityFactory.php
+++ b/src/Spout/Writer/Common/Creator/WriterEntityFactory.php
@@ -2,10 +2,23 @@
 
 namespace Box\Spout\Writer\Common\Creator;
 
+use Box\Spout\Common\Creator\HelperFactory;
 use Box\Spout\Common\Entity\Cell;
 use Box\Spout\Common\Entity\Row;
 use Box\Spout\Common\Entity\Style\Style;
+use Box\Spout\Common\Helper\GlobalFunctionsHelper;
+use Box\Spout\Writer\Common\Creator\Style\StyleBuilder;
+use Box\Spout\Writer\CSV\Manager\OptionsManager as CSVOptionsManager;
+use Box\Spout\Writer\CSV\Writer as CSVWriter;
+use Box\Spout\Writer\ODS\Creator\HelperFactory as ODSHelperFactory;
+use Box\Spout\Writer\ODS\Creator\ManagerFactory as ODSManagerFactory;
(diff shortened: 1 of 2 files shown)


boxcla:
Verified that @sinri has signed the CLA. Thanks for the pull request!

src/Spout/Writer/Common/Creator/WriterFactory.php
```diff
@@ -47,7 +47,7 @@ public function create($writerType)
     /**
      * @return CSVWriter
      */
-    private function getCSVWriter()
+    public function getCSVWriter()
```
adrilo: I would not expose these methods. You don't want people using the lib to have access to them.

sinri: So the duplicated instance create methods would be needed again in Entity Factory?

sinri: Well I have hidden them again.

src/Spout/Writer/Common/Creator/WriterEntityFactory.php
```diff
@@ -25,6 +38,50 @@ public static function createWriter($writerType)
         return (new WriterFactory())->create($writerType);
     }
 
+    /**
+     * @return \Box\Spout\Writer\CSV\Writer
+     */
+    public static function createCSVWriter()
```
adrilo: Why not just calling `return self::createWriter(Type:CSV)`? There's no need to redefine how a CSV writer should be constructed here. That's what I meant with my previous comments.
Also, you'll need to add/update tests.

sinri: Actually the factory design is not good enough when you want to make the PHPDoc correct and let IDE suggest you the methods special for one certain type of writer. Anyway, it is of course able to be realized by manually call the create codes, is not better to provide them rather than a vague factory?

adrilo:
Duplicated here: #642

PR #639, opened 2019-05-17
Update Reader/WriterEntityFactory

adrilo:
- Add `WriterEntityFactory::createWriterFromFile`, working like `ReaderEntityFactory::createReaderFromFile` (guessing writer type from file name).
- Use static functions when needed.


boxcla:

PR #643, opened 2019-05-17
Add methods to SheetInterface

### LINKED ISSUE #626: Add methods to SheetInterface
Hi

As far I can tell, the only thing the files Sheet.php in XLSX/CVS/ODS have in common is the SheetInterface. I'm building a tool that should take a Sheet as input for further processing, no matter which file it is originating from. I'm defining input as of type SheetInterface. It would help a lot, if the following methods are defined in SheetInterface:

    public function getName()
    public function getIndex()
    public function isActive()

Is this possible?

adrilo:
Fixes #626 

The SheetInterface was missing methods common to all Sheets (getIndex, getName, isActive, isVisible).

### DIFF
--- a/src/Spout/Reader/CSV/Sheet.php
+++ b/src/Spout/Reader/CSV/Sheet.php
@@ -51,4 +51,12 @@ class Sheet implements SheetInterface
     {
         return true;
     }
+
+    /**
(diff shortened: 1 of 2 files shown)


boxcla:

PR #644, opened 2019-05-21
Move documentation from gh-pages branch to 'docs' folder

adrilo:
To prepare the migration to 3.0, we need to change the location where the documentation is generated from.
Having a gh-pages branch makes it hard to synchronize the code and the docs. Having a "docs" folder in the repo itself simplifies this.


boxcla:

PR #653, opened 2019-05-29
Adding Fixed Header feature to Excel file

dramosv90:
### Description
Generate new excel files with top header fixed


boxcla:
Verified that @dramosv90 has signed the CLA. Thanks for the pull request!

PR #656, opened 2019-06-03
Fix reading of 1904 dates option

### LINKED ISSUE #655: Wrong date being returned from LibreOffice XLSX: Use1904Date?
I have a large file with a lot of dates, manipulated by both Excel 2013 and this LibreOffice here:

Version: 6.2.3.2 (x64)
Build ID: aecc05fe267cc68dde00352a451aa867b3b546ac
CPU threads: 8; OS: Windows 6.1; UI render: default; VCL: win; 
Locale: it-IT (it_IT); UI-Language: en-US
Calc: threaded

Reading it with Spout, date cells are read in with the wrong date - there is a difference of four years and one day on most, but maybe not all, dates.

By cutting out rows and columns I've boiled it down to a file with a single cell, with a date of  "19/06/2015" (June 19th, 2015 in Italian format), which I am attaching.

This code:

$filePath = 'src/sample.xlsx';
$reader = ReaderEntityFactory::createXLSXReader();
// $reader->setShouldFormatDates(true);

$reader->open($filePath);
/** @var Iterator $sheet */
        /** @var Row $rowObject */
        foreach ($sheet->getRowIterator() as $rowObject) {
            $row = $rowObject->toArray();
            var_dump($row);
        }
}

returns the cell as:

class DateTime#17 (3) {
    public $date =>
    string(26) "2019-06-20 00:00:00.000000"
    public $timezone_type =>
    int(3)
    public $timezone =>
    string(13) "Europe/Berlin"
  }

four years and one day after the desired value. If I set shouldFormatDate to TRUE I get "20/06/2019" as a string (so, still wrong). 

Internally, the number is shown as 42174, which is the correct Excel value for June 19th, 2015.

By checking in the Spout code I found out "$this->shouldUse1904Dates" and discovered that Excel has TWO calendars.

Finally here

      Spout/Reader/XLSX/Manager/SheetManager.php

       protected function processWorkbookPropertiesStartingNode($xmlReader)

I found that Spout reads the XML value of "Should I use 1904 dates?", which is the string "false", and converts it to boolean. But since "false" is a string, and it is not empty, its Boolean equivalent in PHP is TRUE. Therefore, files are read in with the wrong calendar.

I added a more detailed test for falsey values (0, the string "0", the boolean false, the string "false" and "N" just in case) and my file is now decoded correctly.

[sample.xlsx](https://github.com/box/spout/files/3245360/sample.xlsx)

adrilo:
Fixes #655

Whether the spreadsheet is using 1904 dates or not is controlled by a XML property. Its value can be the string "false" that is not mapped to the boolean "false" but to the boolean "true"... Therefore Spout was previously using the wrong date system when this property was set.

### DIFF
--- a/src/Spout/Reader/XLSX/Manager/SheetManager.php
+++ b/src/Spout/Reader/XLSX/Manager/SheetManager.php
@@ -114,7 +114,9 @@ class SheetManager
      */
     protected function processWorkbookPropertiesStartingNode($xmlReader)
     {
-        $shouldUse1904Dates = (bool) $xmlReader->getAttribute(self::XML_ATTRIBUTE_DATE_1904);
+        // Using "filter_var($x, FILTER_VALIDATE_BOOLEAN)" here because the value of the "date1904" attribute
+        // may be the string "false", that is not mapped to the boolean "false" by default...
+        $shouldUse1904Dates = filter_var($xmlReader->getAttribute(self::XML_ATTRIBUTE_DATE_1904), FILTER_VALIDATE_BOOLEAN);
         $this->optionsManager->setOption(Options::SHOULD_USE_1904_DATES, $shouldUse1904Dates);
 
         return XMLProcessor::PROCESSING_CONTINUE;


boxcla:

PR #669, opened 2019-07-20
Support for missing styles XML file in XLSX

### LINKED ISSUE #663: Upgrade v2.7.1 to v3.0.1 : Notice: Undefined index: http://schemas.openxmlformats.org/officeDocument/2006/relationships/styles
Hello,
I have an application with box/spout v2.7.1.
I use it to read an "invalid" xlsx file. Content is valid but there is no "style" sheet.
In this previous version, the xlsx reader was triggering a warning but still read the file. 
Note that I have a warning when I try to read this file with excel too but it works.

After upgrading to v3.0.1 the same process throw an exception:

```
#message: "Notice: Undefined index: http://schemas.openxmlformats.org/officeDocument/2006/relationships/styles"
#code: 0
#file: "./vendor/box/spout/src/Spout/Reader/XLSX/Manager/WorkbookRelationshipsManager.php"
#line: 83
```

Unfortunately the file is provided by a third-party application and it will not be fix.
I'm not able to give you that file because it contains sensitive data, and if I edit it, Excel will fix that "style" issue.

I can help with a pull request if you guide me to the solution you want for this : maybe ignoring the style in that case is enough ?

adrilo:
Fixes #663

Some files don't have a "styles.xml" file. Excel supports these files, Spout should do too.


boxcla:

PR #670, opened 2019-07-20
[PHP 7.4] Updated way to disable the fgetcsv/fputcsv escape character

### LINKED ISSUE #658: PHP 7.4: disable the fgetcsv() escape character
The `GlobalFunctionsHelper::fgetcsv()` wrap sets the escape char to `\0` in order to disable it. With PHP 7.4, it's possible to set it to an empty string in order to disable it.

See https://github.com/box/spout/issues/473

See https://github.com/php/php-src/pull/3515

adrilo:
Fixes #658

From PHP 7.4, the recommended way to disable the escape character for fgetcsv() and fputcsv() is an empty string, instead of "\0".
Discussed here: https://github.com/php/php-src/pull/3515


boxcla:

PR #683, opened 2019-09-19
Cell indexes in Row not being respected when rendering Row

Efiware:
See Issue #682

PR #686, opened 2019-09-28
Cell indexes not being respected when rendering row

### LINKED ISSUE #682: $row->setCellAtIndex doesn't work (as I expect)
Hi

Please have look at this example:

 ```
   $writer = WriterEntityFactory::createXLSXWriter();
    $writer->openToFile("php://output");

    $row = WriterEntityFactory::createRow([]);

    for ($i = 0; $i < 20; $i++) {
      if ($i % 2 == 0) {
        $cell = WriterEntityFactory::createCell('Cell ' . $i);
        $row->setCellAtIndex($cell, $i);
      }
    }
    $writer->addRow($row);
    $writer->close();
```

I would expect the Cell values to be put in every even column, but when I run it, the Cell values are found in the first 10 columns. Am I missing something here, or is this a bug?

adrilo:
Fixes #682
When calling `Row::setCellIndex`, it's possible to create a Row with holes.
Instead of iterating over existing cells of a Row, we should instead use the cell indexes (from 0 to max cell index).

### DIFF
--- a/src/Spout/Common/Entity/Row.php
+++ b/src/Spout/Common/Entity/Row.php
@@ -70,7 +70,7 @@ class Row
      */
     public function getCellAtIndex($cellIndex)
     {
-        return isset($this->cells[$cellIndex]) ? $this->cells[$cellIndex] : null;
+        return $this->cells[$cellIndex] ?? null;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


boxcla:

adrilo:
Inspired by #683.

PR #699, opened 2019-10-27
[ODS] Add support for whitespaces inside <text:span>

### LINKED ISSUE #666: Support for <text:s> inside <text:span> (ODS file)
I have an ODS file with this content (reduced for brevity):

```xml
<table:table-cell office:value-type="string">
  <text:p>
    <text:span>Lorem ipsum<text:s/></text:span>dolor sit amet
  </text:p>
</table:table-cell>
```

I expected this content:

```
Lorem ipsum dolor sit amet
```

But Spout returns this without the space between `ipsum` and `dolor`:

```
Lorem ipsumdolor sit amet
```

Because the `<text:s>` is inside `<text:span>` and `Box\Spout\Reader\ODS\Helper\CellValueFormatter::formatStringCellValue()` works only with direct children of the `p` node.

I suggest to refactor the method by extracting part of the logic to use it recursively. I will make a pull request soon.

adrilo:
Fixes #666

The `<text:p>` node can contain the string value directly or contain child elements. In this case, whitespaces contain in the child elements should be replaced by their XML equivalent:
 - space => `<text:s />`
 - tab => `<text:tab />`
 - line break => `<text:line-break />`

@see https://docs.oasis-open.org/office/v1.2/os/OpenDocument-v1.2-os-part1.html#__RefHeading__1415200_253892949


boxcla:

PR #703, opened 2019-11-16
Add support for cells in error when writing XLSX and ODS

### LINKED ISSUE #701: Add access/documentation to Get and Set cell protected value
I was trying to figure out the issue I am having from issue #204 and #654...  After not finding a solution from either I dug a bit, which wasn't far, it just felt like it because of my lack of knowledge.  And I discovered that my excel sheets occasionally do something dumb and I want to be able to prevent this from creating a fatal error...  My code is as such:
```
foreach($sheet->getRowIterator() as $rowObject){
          $row = $rowObject->getCells();
          for($v = 0; $v < count($row); $v++){
              var_dump($row[$v]);
          }
}
```

Fairly simple.  For the sake of not posting a bunch of useless info I won't show the full object dump.  The important part is here...
`object(Box\Spout\Common\Entity\Cell)#227 (3) { ["value":protected]=> string(7) "#DIV/0!" `

So this wouldn't normally be weird...  In fact why don't I just check the cell value like normal?  So I swapped `var_dump($row[$v]);` with `echo "Value: " . $row[$v];` and this is the result:

`Value: `

So that's inconvenient...

I opened the xlsx sheet in google sheets and did [ctrl + f] for '#DIV' and it found nothing.  Maybe my issue is Google sheets, but previous to this problem I found several formulas which equated to what I would assume to be NULL values, and removing those seemed to solve it.  There is nothing obvious in the data about this particular piece of text...  So Box/Spout is picking up something I can't see or change.  My next solution would be to delete what I assume to be the problem row and try again, which is a bad idea because it is sensitive information, but what else can I do?

Any thoughts, comments, questions, suggestions, would be greatly appreciated...

I was going to try to write a class to extend the Cell class so that I could add Get and Set functions for the protected value but I have no idea how to do that.  Is this even a good idea?...  Thanks for the help.

adrilo:
Fixes #701 

When appending data to an existing sheet, it was possible to get cells in error when reading (DIV/0 for instance). When trying to write them back, `addRow` would throw because `Cell`s in error were not supported by the writers.

### DIFF
--- a/src/Spout/Common/Entity/Cell.php
+++ b/src/Spout/Common/Entity/Cell.php
@@ -92,6 +92,14 @@ class Cell
     }
 
     /**
+     * @return mixed
+     */
+    public function getValueEvenIfError()
+    {
+        return $this->value;
+    }
+
+    /**
      * @param Style|null $style
      */
     public function setStyle($style)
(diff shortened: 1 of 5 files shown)


boxcla:

PR #704, opened 2019-11-17
disable no_superfluous_phpdoc_tags

madflow:
* Disables "no_superfluous_phpdoc_tags"
* See https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2979

* Other options: Enable it and fix all occurences
* Pin Cs Fixer to a specific version, where this option was not merged in upstream


boxcla:

PR #705, opened 2019-11-24
Enable support to XLSM spreadsheets

marcosteodoro:
I needed to read some XLSM files with this package, but the extension weren't match with the file types supported. Testing with the XLSX reader, I saw that the package is able to parse the information that I needed.

I made this changes and tests, and worked fine for me.

To tell the truth, I don't know if this run away of the project scope, but it helped me a lot, and is my way to say thank you!


boxcla:
Verified that @marcosteodoro has signed the CLA. Thanks for the pull request!

adrilo:
Thanks for the contribution @marcosteodoro !

Adding support for a "new" format is a big commitment. If we add support for XLSM, I would expect some kind of support for the features XLSM offers on top of the XLSX features. If their behavior diverge at one point, then you now need to support both separately, etc..

For these reasons, I'm not going to accept this PR but that remains a great source for people who are looking for adding XLSM support in their project. Thanks again!

PR #712, opened 2019-12-14
add PHP 7.4

madflow:
* Add PHP 7.4 to the test grid
* Fix "PHP Deprecated:  strpos(): Non-string needles will be interpreted as strings in the future. Use an explicit chr() call to preserve the current behavior in /home/travis/build/madflow/spout/src/Spout/Common/Helper/FileSystemHelper.php on line 126"
* `FileSystemHelper::baseFolderRealPath` is set with `realpath` - which can also be `false`.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -8,6 +8,10 @@
         defaultTestSuite="unit-tests"
         verbose="false">
 
+    <php>
+        <ini name="error_reporting" value="-1" />
+    </php>
+
     <testsuites>
 
         <testsuite name="unit-tests">
--- a/src/Spout/Common/Helper/FileSystemHelper.php
+++ b/src/Spout/Common/Helper/FileSystemHelper.php
@@ -17,7 +17,7 @@ class FileSystemHelper implements FileSystemHelperInterface
     /**
      * @param string $baseFolderPath The path of the base folder where all the I/O can occur
      */
-    public function __construct($baseFolderPath)
+    public function __construct(string $baseFolderPath)
     {
         $this->baseFolderRealPath = \realpath($baseFolderPath);
     }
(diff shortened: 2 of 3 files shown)


boxcla:

src/Spout/Common/Helper/FileSystemHelper.php
```diff
@@ -117,12 +117,16 @@ public function deleteFolderRecursively($folderPath)
...
-     * @throws \Box\Spout\Common\Exception\IOException If the folder where the I/O operation should occur is not inside the base folder
+     * @throws \Box\Spout\Common\Exception\IOException If the folder where the I/O operation should occur
+     * is not inside the base folder or the base folder does not exist
      * @return void
      */
-    protected function throwIfOperationNotInBaseFolder($operationFolderPath)
+    protected function throwIfOperationNotInBaseFolder(string $operationFolderPath)
     {
         $operationFolderRealPath = \realpath($operationFolderPath);
+        if (!$this->baseFolderRealPath) {
```
adrilo: should this check be in the constructor instead? If `baseFolderRealPath` is `false`, it means that `realpath` call in the ctor failed; so we can catch the error there directly?

madflow: This would also work with a small adjustments of the tests. The tests rely on a non existing base folder - maybe a user does, too? I could not think of an use case, though.

madflow:
* I also set the error_reporting to `-1` for the unit tests. This was implied on the Travis host. This change should make deprecation warning more visible (it had been there since PHP 7.3).

src/Spout/Common/Helper/FileSystemHelper.php
adrilo: Anyway, it's probably not worth making the change. The program fails pretty quickly with the current implementation.

phpunit.xml
```diff
@@ -8,6 +8,10 @@
         defaultTestSuite="unit-tests"
         verbose="false">
 
+    <php>
+        <ini name="error_reporting" value="-1" />
```
adrilo: 👍

PR #714, opened 2019-12-18
use existing base folder

madflow:
* in #712 I created a false positive test
* The actual IOException thrown was not the one the tests covered - but the newly added check for an existing base folder
* here I simply use `\sys_get_temp_dir();` for a hopefully existing folder and check the Exception messages


boxcla:

PR #717, opened 2020-01-06
Allow accent in Filename

bouland:
Protected function allow overwrite without full copy paste
Even if you have to rewrite the factory method anyway ...

first filename will be a fallback in ASCII for browser not supporting second filename with asterisk

### DIFF
--- a/src/Spout/Writer/WriterAbstract.php
+++ b/src/Spout/Writer/WriterAbstract.php
@@ -125,7 +125,7 @@ abstract class WriterAbstract implements WriterInterface
 
         // Set headers
         $this->globalFunctionsHelper->header('Content-Type: ' . static::$headerContentType);
-        $this->globalFunctionsHelper->header('Content-Disposition: attachment; filename="' . $this->outputFilePath . '"');
+        $this->globalFunctionsHelper->header($this->getContentDispositionHeader());
 
         /*
          * When forcing the download of a file over SSL,IE8 and lower browsers fail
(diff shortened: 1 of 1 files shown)


boxcla:
Hi @bouland, thanks for the pull request. Before we can merge it, we need you to sign our Contributor License Agreement. You can do so electronically here: http://opensource.box.com/cla

bouland:
CLA signed

boxcla:
Verified that @bouland has just signed the CLA. Thanks, and we look forward to your contribution.

adrilo:
@bouland can you please give more details on why this is needed?
I was able to create a file with an accent without your change. Which problem does it solve?

bouland:
@adrilo how ? did u try an end to end test ? with a browser i mean.

Header accept only ASCII chars and us-ascii do not support accent. have a look to RFC i have mention in comments.

I had transcoded chars in my downloads, i have seen that in Network Tab DevTools, maybe it was the browser who perfom the encoding,

This PR fixed my usage on my project. (I have overwrited the Writer with Symfony)

adrilo:
I used this code:
```
$writer = WriterEntityFactory::createXLSXWriter();
$writer->openToBrowser('état.xlsx');

$rowFromValues = WriterEntityFactory::createRowFromArray(['dolor sit amet']);
$writer->addRow($rowFromValues);
$writer->close();
```

I can confirm that the generated file is called "état.xlsx".

PR #744, opened 2020-05-15
feat: add sheet view support to XLSX writer

kjkooistra-youwe:
This makes it possible to freeze rows and columns, in addition to providing
other sheet view settings like zoom options.


Slamdunk:
Merged in https://github.com/openspout/openspout/pull/6

PR #747, opened 2020-05-29
Fixed code style (probably due to recent php-cs-fixer version)

jmsche:
Hello,

This PR fixes code style issues that are triggered in Travis, probably due to a recent release of php-cs-fixer.

PR #748, opened 2020-05-29
Shorter (relevant) diff by php-cs-fixer for Travis CI

jmsche:
Hello,

This PR makes the report of php-cs-fixer way shorter by printing only the diff parts. The current diff format prints the diff parts and the rest of the file, which makes it pretty hard to browse.

Travis failure will be unrelated (see #747)

Update: you can compare [this report (with shorter diff)](https://travis-ci.org/github/box/spout/jobs/692493395) and [this one (with current diff format)](https://travis-ci.org/github/box/spout/jobs/692494711) to see the difference :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ before_script:
 script:
   - |
     if [[ "$WITH_CS" == "true" ]]; then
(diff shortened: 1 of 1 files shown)


adrilo:
Much better indeed. Thanks!

Issue #751, opened 2020-06-08
Writer doesn't support writing `DateTime` objects


guss77:
According to `Box\Spout\Common\Entity\Cell`, a `DateTime` object is a valid cell value, and it is correctly assigned the type `TYPE_DATE`. But it seems the writer implementations do not support it: when trying to write a sheet that contains a `TYPE_DATE` cell, the writer throws `Box\Spout\Common\Exception\InvalidArgumentException: Trying to add a value with an unsupported type: object`

(Happened in ODS writer but the code looks similar for XLS and looks like CSV just tries to stringify all the cells, which will not work because `DateTime` does not have a `__toString()` method).

ChronicLogic:
I added a PR #781  for XLSX, I had addressed this separately, unaware that there was an issue with ODS or CSV.  Whoever (if anyone) would integrate the PRs should look at all of the cases.

adrilo:
Duplicates #662

PR #752, opened 2020-06-08
ODS Writer support for 'date' and 'time' types

guss77:
Fixes issue #751

Should probably be ported to the XLS writer as well, but I don't have the time or inclination to learn the XLS format ATM.

CLA was signed by me.


guss77:
BTW - there's a bug in LibreOffice when reading date fields: if your `DateTime` object is not in UTC, then Calc will read the time correctly, convert it to UTC, drop the timezone information and then show you the UTC-formatted time regardless of which time zone your PC is in.

This is documented in this bug report: https://bugs.documentfoundation.org/show_bug.cgi?id=121851

guss77:
Closing due to #765 superseding this.

BTW, regarding the previous comment, according to spec, the written value should be UTC or non-timezones, so the comment is incorrect as well as the bug it points to.

PR #763, opened 2020-08-25
XLSX and ODS write optimization

alamirault:
This PR try to fix XLSX and ODS write performances due to Styles (#624)

- [x] `EmptyStyle` in order to avoid row merging

- [x] Avoid serialization twice during registerStyle

- [x] Register style can be skipped when already registered

With these 3 changes on 1000 row of 3 cells: **40% less**
On 1000 row of 9 cells: **49% less**

image

## Break changes:

Methods `applyWrapTextIfCellContainsNewLine` and `applyExtraStylesIfNeeded` of `StyleManager` return is now `null` when no extra style is applied.

### DIFF
--- a/src/Spout/Common/Entity/Style/Style.php
+++ b/src/Spout/Common/Entity/Style/Style.php
@@ -84,6 +84,9 @@ class Style
     /** @var bool */
     private $hasSetFormat = false;
 
+    /** @var bool */
+    private $isRegistered = false;
+
     /**
      * @return int|null
      */
@@ -462,5 +465,24 @@ class Style
     public function shouldApplyFormat()
     {
         return $this->hasSetFormat;
+    }
+
+    public function setRegistered(bool $isRegistered = true) : void
+    {
+        $this->isRegistered = $isRegistered;
+    }
+
+    /**
+     * @return bool
+     */
+    public function isRegistered() : bool
+    {
+        return $this->isRegistered;
+    }
+
+    public function register(int $id) : void
+    {
+        $this->setId($id);
+        $this->isRegistered = true;
     }
 }
--- a/src/Spout/Common/Entity/Style/Style.php
+++ b/src/Spout/Common/Entity/Style/Style.php
@@ -87,6 +87,9 @@ class Style
     /** @var bool */
     private $isRegistered = false;
 
+    /** @var bool */
+    private $isEmpty = true;
+
     /**
      * @return int|null
      */
@@ -122,6 +125,7 @@ class Style
     {
         $this->shouldApplyBorder = true;
         $this->border = $border;
+        $this->isEmpty = false;
 
         return $this;
     }
@@ -150,6 +154,7 @@ class Style
         $this->fontBold = true;
         $this->hasSetFontBold = true;
         $this->shouldApplyFont = true;
+        $this->isEmpty = false;
 
         return $this;
     }
@@ -178,6 +183,7 @@ class Style
         $this->fontItalic = true;
         $this->hasSetFontItalic = true;
         $this->shouldApplyFont = true;
+        $this->isEmpty = false;
 
         return $this;
     }
@@ -206,6 +212,7 @@ class Style
         $this->fontUnderline = true;
         $this->hasSetFontUnderline = true;
         $this->shouldApplyFont = true;
+        $this->isEmpty = false;
 
         return $this;
     }
@@ -234,6 +241,7 @@ class Style
         $this->fontStrikethrough = true;
         $this->hasSetFontStrikethrough = true;
         $this->shouldApplyFont = true;
+        $this->isEmpty = false;
 
         return $this;
     }
@@ -263,6 +271,7 @@ class Style
         $this->fontSize = $fontSize;
         $this->hasSetFontSize = true;
         $this->shouldApplyFont = true;
+        $this->isEmpty = false;
 
         return $this;
     }
@@ -294,6 +303,7 @@ class Style
         $this->fontColor = $fontColor;
         $this->hasSetFontColor = true;
         $this->shouldApplyFont = true;
+        $this->isEmpty = false;
 
         return $this;
     }
@@ -323,6 +333,7 @@ class Style
         $this->fontName = $fontName;
         $this->hasSetFontName = true;
         $this->shouldApplyFont = true;
+        $this->isEmpty = false;
 
         return $this;
     }
@@ -353,6 +364,7 @@ class Style
         $this->cellAlignment = $cellAlignment;
         $this->hasSetCellAlignment = true;
         $this->shouldApplyCellAlignment = true;
+        $this->isEmpty = false;
 
         return $this;
     }
@@ -389,6 +401,7 @@ class Style
     {
         $this->shouldWrapText = $shouldWrap;
         $this->hasSetWrapText = true;
+        $this->isEmpty = false;
 
         return $this;
     }
@@ -418,6 +431,7 @@ class Style
     {
         $this->hasSetBackgroundColor = true;
         $this->backgroundColor = $color;
+        $this->isEmpty = false;
 
         return $this;
     }
@@ -447,6 +461,7 @@ class Style
     {
         $this->hasSetFormat = true;
         $this->format = $format;
+        $this->isEmpty = false;
 
         return $this;
     }
@@ -467,11 +482,6 @@ class Style
         return $this->hasSetFormat;
     }
 
-    public function setRegistered(bool $isRegistered = true) : void
-    {
-        $this->isRegistered = $isRegistered;
-    }
-
     /**
      * @return bool
      */
@@ -484,5 +494,16 @@ class Style
     {
         $this->setId($id);
         $this->isRegistered = true;
+    }
+
+    public function unregister() : void
+    {
+        $this->setId(0);
+        $this->isRegistered = false;
+    }
+
+    public function isEmpty() : bool
+    {
+        return $this->isEmpty;
     }
 }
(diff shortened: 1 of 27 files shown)


alamirault:
PR is now ready ! :smiley:

devnix:
Any plans to merge this PR?

adrilo (changes requested):

src/Spout/Writer/Common/Manager/Style/StyleRegistry.php
```diff
@@ -57,6 +57,17 @@ protected function hasStyleAlreadyBeenRegistered(Style $style)
     {
```
adrilo, line 57: This method is no longer relevant, as we have the serialized style everywhere we call it.
I'd remove it and only keep your new method.

src/Spout/Writer/ODS/Manager/WorksheetManager.php
```diff
@@ -155,14 +155,32 @@ public function addRow(Worksheet $worksheet, Row $row)
      * @throws InvalidArgumentException If a cell value's type is not supported
      * @return string
      */
-    private function applyStyleAndGetCellXML(Cell $cell, Style $rowStyle, $currentCellIndex, $nextCellIndex)
+    private function applyStyleAndGetCellXML(Cell $cell, Style &$rowStyle, $currentCellIndex, $nextCellIndex)
```
adrilo, line 158: I'm not a fan of using reference passing. I see this as a side effet that can easily be missed or go out of control.
This also is a good sign that this method does too many things and it's a great time to refactor it. Maybe compute the rowStyle first, then the cell style, then return the XML?

src/Spout/Writer/ODS/Manager/WorksheetManager.php
```diff
@@ -155,14 +155,32 @@ public function addRow(Worksheet $worksheet, Row $row)
...
-        $newCellStyle = $this->styleManager->applyExtraStylesIfNeeded($cell);
+        if ($cell->getStyle()->isEmpty()) {
+            $cell->setStyle($rowStyle);
+
+            $managedStyle = $this->styleManager->applyExtraStylesIfNeeded($cell);
+
+            if ($managedStyle->isUpdated()) {
+                $registeredStyle = $this->styleManager->registerStyle($managedStyle->getStyle());
+            } else {
+                $registeredStyle = $rowStyle = $this->styleManager->registerStyle($rowStyle);
```
adrilo, line 168: split in two lines to improve readability

src/Spout/Writer/XLSX/Manager/WorksheetManager.php
```diff
@@ -183,14 +183,32 @@ private function addNonEmptyRow(Worksheet $worksheet, Row $row)
...
+        if ($cell->getStyle()->isEmpty()) {
+            $cell->setStyle($rowStyle);
 
-        $registeredStyle = $this->styleManager->registerStyle($newCellStyle);
+            $managedStyle = $this->styleManager->applyExtraStylesIfNeeded($cell);
+
+            if ($managedStyle->isUpdated()) {
+                $registeredStyle = $this->styleManager->registerStyle($managedStyle->getStyle());
+            } else {
+                $registeredStyle = $rowStyle = $this->styleManager->registerStyle($rowStyle);
```
adrilo, line 196: split in 2 lines

src/Spout/Writer/XLSX/Manager/WorksheetManager.php
```diff
@@ -183,14 +183,32 @@ private function addNonEmptyRow(Worksheet $worksheet, Row $row)
      * @throws InvalidArgumentException If the given value cannot be processed
      * @return string
      */
-    private function applyStyleAndGetCellXML(Cell $cell, Style $rowStyle, $rowIndexOneBased, $columnIndexZeroBased)
+    private function applyStyleAndGetCellXML(Cell $cell, Style &$rowStyle, $rowIndexOneBased, $columnIndexZeroBased)
```
adrilo, line 186: same here about the reference

src/Spout/Writer/ODS/Manager/WorksheetManager.php
alamirault: Indeed I can avoid reference, if we register rowStyle first in `addNonEmptyRow`, before calling `applyStyleAndGetCellXML`.
But register rowStyle even if is not used is not a bad thing ?

src/Spout/Common/Entity/Style/Style.php
```diff
@@ -490,13 +490,13 @@ public function isRegistered() : bool
         return $this->isRegistered;
     }
 
-    public function register(int $id) : void
+    public function markAsRegistered(?int $id) : void
```
adrilo, line 493: can this take a `null`?

src/Spout/Writer/ODS/Manager/WorksheetManager.php
adrilo: Well, registering a style increases the memory consumption. If the styles being registered as always or almost always the same, then the impact on the memory is limited so that could be a valid option.

Another option is to split what's done here in 2:
1. Compute and register the cell/row style
2. Get the XML for the registered style

src/Spout/Common/Entity/Style/Style.php
alamirault: Yes, Style id from StyleBuilder is null and it can be passed in construct of `StyleRegistry` for example

src/Spout/Writer/ODS/Manager/WorksheetManager.php
alamirault: I created `RegisteredStyle` like `ManagedStyle` in order to avoid reference.  I also split `applyStyleAndGetCellXML` to `applyStyleAndRegister` and `getCellXMLWithStyle`

src/Spout/Writer/Common/Manager/RegisteredStyle.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Box\Spout\Writer\Common\Manager;
+
+use Box\Spout\Common\Entity\Style\Style;
+
+class RegisteredStyle
```
adrilo, line 7: I think it's worth documenting what the style is about (as the logic is not trivial to understand)

src/Spout/Writer/Common/Manager/Style/ManagedStyle.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace Box\Spout\Writer\Common\Manager\Style;
+
+use Box\Spout\Common\Entity\Style\Style;
+
+class ManagedStyle
```
adrilo, line 7: I think it's worth documenting what the style is about (as the logic is not trivial to understand)

adrilo: Also, maybe rename that `PossiblyUpdatedStyle` or something along these lines.

src/Spout/Writer/Common/Manager/RegisteredStyle.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * @var Style
+     */
+    private $style;
+
+    /**
+     * @var bool
+     */
+    private $isRowStyle;
```
adrilo, line 17: isRowStyle => isMatchingRowStyle

src/Spout/Writer/ODS/Manager/WorksheetManager.php
```diff
@@ -125,7 +126,13 @@ public function addRow(Worksheet $worksheet, Row $row)
             $nextCell = isset($cells[$nextCellIndex]) ? $cells[$nextCellIndex] : null;
 
             if ($nextCell === null || $cell->getValue() !== $nextCell->getValue()) {
-                $data .= $this->applyStyleAndGetCellXML($cell, $rowStyle, $currentCellIndex, $nextCellIndex);
+                $registeredStyle = $this->applyStyleAndRegister($cell, $rowStyle);
+                $cellStyle = $registeredStyle->getStyle();
+                if ($registeredStyle->isRowStyle()) {
```
adrilo, line 131: I don't understand this `if`. `$registerStyle->isRowStyle()` returns true when cell style is empty (and we therefore use the row style). So isn't it already what the `$rowStyle = $row->getStyle();` returns?

Maybe worth adding a comment here to explain the logic

src/Spout/Writer/XLSX/Manager/WorksheetManager.php
```diff
@@ -160,7 +161,12 @@ private function addNonEmptyRow(Worksheet $worksheet, Row $row)
         $rowXML = '<row r="' . $rowIndexOneBased . '" spans="1:' . $numCells . '">';
 
         foreach ($row->getCells() as $columnIndexZeroBased => $cell) {
-            $rowXML .= $this->applyStyleAndGetCellXML($cell, $rowStyle, $rowIndexOneBased, $columnIndexZeroBased);
+            $registeredStyle = $this->applyStyleAndRegister($cell, $rowStyle);
+            $cellStyle = $registeredStyle->getStyle();
+            if ($registeredStyle->isRowStyle()) {
```
adrilo, line 166: same here, worth adding a comment

adrilo (commented):
It's looking much better now! Thanks @alamirault 
Only a few minor things remaining :)

alamirault:
I think is now good, I hope it's ok for you :slightly_smiling_face:

adrilo (approved):

adrilo:
Appreciate your contribution! Merci :)

PR #768, opened 2020-09-16
switch to Github actions

madflow:
* Travis needs owner intervention
* Removed Scrutinizer
* Set `phpdoc_var_without_name` to `true` - so this is not fixed for the 10th time


madflow:
* https://github.com/madflow/spout/actions/runs/257972374

adrilo:
You're missing scrutinizer integration :) It finds some interesting things from time to time!

madflow:
> You're missing scrutinizer integration :) It finds some interesting things from time to time!

I removed this intentionally. I reckon this could just more thing that requires repo owners intervention to function properly in the future. I can see the value it provides.

PR #778, opened 2020-11-29
Fix private methods cannot be final as they are never overridden by other classes

dhonnabhain:
This PR fix the `Private methods cannot be final as they are never overridden by other classes` warning and the `During class fetch: Uncaught The magic method Box\Spout\Common\Singleton::__wakeup() must have public visibility` on php 8.0.


adrilo:
Branch 2.x is considered "closed".

PR #781, opened 2020-12-12
Add Writer support for DateTime values (reader already supports and returns this type)

ChronicLogic:
Simple change I think.


adrilo:
I don't think it's that easy. Here you're formatting the date as a string (ATOM format). For XLSX, dates should be formatted as a timestamp and a custom date style. Not as trivial as one might think!

Slamdunk:
Solved in https://github.com/openspout/openspout/pull/13

PR #784, opened 2020-12-28
Add full support for PHP 8.0

skodak:
Unfortunately due to PHPUnit 8.5 dependency this also drops support for PHP 7.1.

I did not know what version to use due to the bump in minimal PHP version, so I guessed it would be 3.2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /vendor
 /composer.lock
 /.php_cs.cache
+/.phpunit.result.cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,15 +5,15 @@ language: php
 
 matrix:
   include:
-    - php: 7.1
-      env: WITH_CS=true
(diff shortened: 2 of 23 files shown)


skodak:
This is not a real pull request, I just wanted to see how much work it would be. I understand that dropping support for PHP 7.1 and 7.2 might make some people a bit angry.

adrilo (approved):

adrilo:
Thanks @skodak ! I think it's fair to deprecate PHP 7.1 (EOL was 1+ year ago: https://www.php.net/supported-versions.php)

jmsche:
@adrilo Could you please tag 3.2.0 soon? Or do you need other PRs to be fixed/merged before?

jmsche:
@adrilo friendly ping :)

adrilo:
Done :)

jmsche:
Thank you :)

Issue #792, opened 2021-02-07
Boolean cell values in ODS export are created incorrectly.


guss77:
According to the Open Document specification for [`office:boolean-value`](http://docs.oasis-open.org/office/v1.2/os/OpenDocument-v1.2-os-part1.html#__RefHeading__1417624_253892949) and [`datatype boolean`](http://docs.oasis-open.org/office/v1.2/os/OpenDocument-v1.2-os-part1.html#datatype-boolean), the boolean data value should be created as `"true"` or `"false"`. Instead, Spout just writes out the PHP boolean value's conversion to string, which is `"1"` for `true` and `""` (the empty string) for `false`.

This, BTW, works surprisingly well for known good implementations of ODS, such as Libre Office and Google Docs, but - of course - breaks Microsoft ODS import filter, which loads everything as `TRUE` for some reason.

guss77:
PR merged

PR #793, opened 2021-02-07
write boolean value according to spec

guss77:
fixes issue #792

### DIFF
--- a/src/Spout/Writer/ODS/Manager/WorksheetManager.php
+++ b/src/Spout/Writer/ODS/Manager/WorksheetManager.php
@@ -197,7 +197,7 @@ class WorksheetManager implements WorksheetManagerInterface
 
             $data .= '</table:table-cell>';
         } elseif ($cell->isBoolean()) {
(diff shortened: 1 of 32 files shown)


src/Spout/Writer/ODS/Manager/WorksheetManager.php
```diff
@@ -197,7 +197,8 @@ private function getCellXML(Cell $cell, $styleIndex, $numTimesValueRepeated)
 
             $data .= '</table:table-cell>';
         } elseif ($cell->isBoolean()) {
-            $data .= ' office:value-type="boolean" calcext:value-type="boolean" office:boolean-value="' . $cell->getValue() . '">';
+            $value = $cell->getValue() ? 'true' : 'false';
```
adrilo, line 200: can you add a comment with the link to the spec? https://docs.oasis-open.org/office/v1.2/os/OpenDocument-v1.2-os-part1.html#datatype-boolean

guss77: done.

Issue #797, opened 2021-02-24
Date/datetime cell type not implemented for insert?


devnix:
I'm trying to insert a DateTime into a XLSX with no luck, this exception gets thrown: `Trying to add a value with an unsupported type: object`

```php
    private function getCellXML($rowIndexOneBased, $columnIndexZeroBased, Cell $cell, $styleId)
    {
        $columnLetters = CellHelper::getColumnLettersFromColumnIndex($columnIndexZeroBased);
        $cellXML = '<c r="' . $columnLetters . $rowIndexOneBased . '"';
        $cellXML .= ' s="' . $styleId . '"';

        if ($cell->isString()) {
            $cellXML .= $this->getCellXMLFragmentForNonEmptyString($cell->getValue());
        } elseif ($cell->isBoolean()) {
            $cellXML .= ' t="b"><v>' . (int) ($cell->getValue()) . '</v></c>';
        } elseif ($cell->isNumeric()) {
            $cellXML .= '><v>' . $cell->getValue() . '</v></c>';
        } elseif ($cell->isError() && is_string($cell->getValueEvenIfError())) {
            // only writes the error value if it's a string
            $cellXML .= ' t="e"><v>' . $cell->getValueEvenIfError() . '</v></c>';
        } elseif ($cell->isEmpty()) {
            if ($this->styleManager->shouldApplyStyleOnEmptyCell($styleId)) {
                $cellXML .= '/>';
            } else {
                // don't write empty cells that do no need styling
                // NOTE: not appending to $cellXML is the right behavior!!
                $cellXML = '';
            }
        } else {
            throw new InvalidArgumentException('Trying to add a value with an unsupported type: ' . \gettype($cell->getValue()));
        }

        return $cellXML;
    }
```

It's me, or the DateTime cell is not implemented in `Box\Spout\Writer\XLSX\Manager\WorksheetManager::getCellXML()`?

adrilo:
Hello !
It's not implemented indeed. It's actually tricky to implement correctly. In XLSX, a date is composed of a timestamp + a custom style representing how it's formatted. There's no plan to add this feature in the short term unfortunately.

devnix:
@adrilo Borrowing the implementation from PhpSpreadsheet would slow down Spout? I know there is a list of edge cases, like replicating bugs for some leap years that really aren't and things like that. Would you accept that kind of contribution?

adrilo:
[...]

PR #816, opened 2021-05-24
XLSX Date Writer

rodnaph:
Adds support for writing instances of `DateTimeInterface` (ie. `DateTime` and `DateTimeImmutable`) with the XLSX writer.

Previous issues such as #781 and #797 have been closed with regards this issue, but I believe #781 is actually valid. From my research (disclaimer: I'm new to this XML format...) it seems the numeric date format for "serial dates" referenced in the closing comment of #781 are those originally supported by openxml, for example...

```xml
<c s="14">
  <v>44401</v>
</c>
```

However, ISO/IEC 29500-1:2016 adds support for ISO dates using the type `d`.

https://www.iso.org/standard/71691.html

image

```xml
<c t="d" s="14">
  <v>2021-02-03T12:34:45</v>
</c>
```

Thoughts? If this is valid I'm happy to re-work this PR as needed, cheers.

### DIFF
--- a/src/Spout/Common/Helper/CellTypeHelper.php
+++ b/src/Spout/Common/Helper/CellTypeHelper.php
@@ -61,7 +61,7 @@ class CellTypeHelper
     public static function isDateTimeOrDateInterval($value)
     {
         return (
-            $value instanceof \DateTime ||
+            $value instanceof \DateTimeInterface ||
             $value instanceof \DateInterval
         );
     }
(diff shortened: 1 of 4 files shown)


rodnaph:
I tested Numbers on Mac 6.2 and it does support this format, I'm surprised Google Sheets doesn't though...  I'll investigate and re-open this if I can find a way around it.

rodnaph:
Would you accept a PR that omitted the `t="d"` attribute and instead wrote the `DateTimeInterface` as a serial date? (eg. `40021`)

adrilo:
Using a timestamp would be the best approach indeed. But it requires setting a cell style (there are some built-in styles for dates that can be used). This also poses some questions:
- should you let the user choose the date format?
- If not, which format should you choose? Is this format localized?

Slamdunk:
Solved in https://github.com/openspout/openspout/pull/13

PR #822, opened 2021-06-22
[DX] more helpful exception message in case of invalid value

andrew-demb:
It is should provide a more helpful message for developer in exception, when in cell value provided an object.

**Example**:

Exception:
```
Uncaught PHP Exception Box\\Spout\\Common\\Exception\\InvalidArgumentException: \"Trying to add a value with an unsupported type: NULL\" at /project/vendor/box/spout/src/Spout/Writer/XLSX/Manager/WorksheetManager.php line 256
```

I may assume, that in my code somewhere used `null` and passed to `spout`, but the real reason was - object was provided for cell value, and this lead to described above exception.

This PR will improve an error message and provide a better DX.

### DIFF
--- a/src/Spout/Writer/XLSX/Manager/WorksheetManager.php
+++ b/src/Spout/Writer/XLSX/Manager/WorksheetManager.php
@@ -253,7 +253,7 @@ EOD;
                 $cellXML = '';
             }
         } else {
-            $value = $cell->getValue();
+            $value = $cell->getValueEvenIfError();
 
             throw new InvalidArgumentException('Trying to add a value with an unsupported type: ' . (\is_object($value) ? \get_class($value) : \gettype($value)));
         }
--- a/src/Spout/Writer/XLSX/Manager/WorksheetManager.php
+++ b/src/Spout/Writer/XLSX/Manager/WorksheetManager.php
@@ -253,7 +253,9 @@ EOD;
                 $cellXML = '';
             }
         } else {
-            throw new InvalidArgumentException('Trying to add a value with an unsupported type: ' . \gettype($cell->getValue()));
+            $value = $cell->getValue();
+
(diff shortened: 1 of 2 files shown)


adrilo:
Hi!
I believe that your interpretation may not be correct. `gettype($object)` should return "object" and not "NULL" like you describe. Only `gettype(null)` returns "NULL".
So here, it seems like a NULL value is passed in.

andrew-demb:
Yes, it was output NULL because `$cell->getValue()` returned `null`, but this method return null, because resolved cell type was an `error`

So PR propose change `$cell->getValue()` to `$cell->getValueEvenIfError()` AND explain object class name, when it is contained in value.

src/Spout/Writer/XLSX/Manager/WorksheetManager.php
```diff
@@ -253,7 +253,9 @@ private function getCellXML($rowIndexOneBased, $columnIndexZeroBased, Cell $cell
                 $cellXML = '';
             }
         } else {
-            throw new InvalidArgumentException('Trying to add a value with an unsupported type: ' . \gettype($cell->getValue()));
+            $value = $cell->getValueEvenIfError();
+
+            throw new InvalidArgumentException('Trying to add a value with an unsupported type: ' . (\is_object($value) ? \get_class($value) : \gettype($value)));
```
adrilo, line 258: I think keeping "object" is clearer here. I'd suggest a combination of both:
```suggestion
$errorMessage = 'Trying to add a value with an unsupported type: ' . \gettype($value);
if (is_object($value)) {
    $errorMessage .= ' (' . get_class($value) . ' given)';
}

throw new InvalidArgumentException($errorMessage);
```

WDYT?

andrew-demb: I'm fine with that.
The previous implementation was inspired by Symfony codebase - they show only native or classname.

andrew-demb:
Probably Travis CI check should be retried because PR check status was not changed to finished status.

PR #839, opened 2021-11-10
Add dedicated RowIteratorInterface and SheetIteratorInterface

jmleroux:
Dedicated interfaces are needed for correct type-hinting.

When using PHP >= 7.4, a more strict interface would for example allow safe use of Row::toArray() when using the iterator.
If we keep the generic Iterator interface, the retruned item is `mixed` type and the use of `$item->toArray()` is not safe.


Slamdunk:
Merged in https://github.com/openspout/openspout/pull/7

PR #840, opened 2021-11-10
Add Docker files and documentation

jmleroux:
I didn't have PHP 7.2 on my machine and had to use Docker to run unit tests

PR #841, opened 2021-11-12
Setup Github actions

adrilo:


jmleroux:
Great initiative!

jmleroux:
FYI, it could also work with Docker: https://github.com/box/spout/pull/840

Issue #844, opened 2021-12-10
PHP 8.1 Issue with basename


jansor:
Hi, i was testing some items against PHP8.1

when export to browser it make use of basename method().

basename(): Passing null to parameter #2 ($suffix) of type string is deprecated
box\spout\src\Spout\Common\Helper\GlobalFunctionsHelper.php:251

I hope in the next release this could be fixed.
Thank you for the amazing library, greet job

PR #859, opened 2022-01-25
Update GlobalFunctionsHelper.php

joergmoenke:
PHP 8.1 issue. $suffix may not be null when using PHP's basename function. Default value is an empty string. Belongs to issue #844

### DIFF
--- a/src/Spout/Common/Helper/GlobalFunctionsHelper.php
+++ b/src/Spout/Common/Helper/GlobalFunctionsHelper.php
@@ -248,7 +248,7 @@ class GlobalFunctionsHelper
      */
     public function basename($path, $suffix = null)
     {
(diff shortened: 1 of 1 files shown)


src/Spout/Common/Helper/GlobalFunctionsHelper.php
```diff
@@ -248,7 +248,7 @@ public function is_readable($fileName)
      */
     public function basename($path, $suffix = null)
```
adrilo, line 249: looking at the method signature (https://www.php.net/manual/en/function.basename), it should be:
```
public function basename($path, $suffix = '')
```
Changing the signature (and associated PHPDoc) would be a better fix I believe.

Issue #129, opened 2015-10-22
Automatically calculate columns width / rows height


adrilo:
Instead of using the default width/height, Spout could calculate the optimal width/height for each cell, based on its content.

This involves using a graphics library, render the text content with the chosen font and get the width/height of the rendered text.

This calculation may be slow so it should be optional. It can also be used when using "wrap text" as wrapping does not work well without a custom row height in LibreOffice.

zendre4:
+1

codewizz:
+1

tomikulmala:
+1 Would be a really good feature!

arisdario:
+10

jordanade:
I don't need auto-fit, just a way to set widths for each column! Any way to workaround/hack this for now?

jordanade:
Oh, figured it out--just added this line to Worksheet.php, startSheet():
`fwrite($this->sheetFilePointer, '<cols><col min="1" max="1" width="100" customWidth="1"/><col min="2" max="6" width="30" customWidth="1"/></cols>');`

adrilo:
@jordanade You can also look at #225 if you need to hack something that works.
I'm not a big fan of this as it creates an API that is specific to XLSX. Also, if Spout could automatically calculate the width of each column, it would solve 95% of the problems.

jordanade:
Surely there is room for APIs that may not work for all possible file outputs? Best effort, etc...

adrilo:
I'm actually trying to avoid this as much as possible. Although I'll definitely reconsider after implementing the automated method.

nimmneun:
Hi there, have you been able to come up with a solution for the autosizing of column widths and row heights already?

If not, I've come with an intermediary solution, which - while less accurate than imagettfbbox - is IMHO a decent tradeoff between performance and accuracy.
On average it increases the runtime by about 50%-100% and the widths are on average about 1 character off (usually too wide).

It's a bit hacky though (requires serious refinement! =) and involves a *.csv that holds base sizes for each font and character in the ASCII range, which is why I'm hesitating to fork->code->pull spout, since I dont know if it's a viable solution to integrate that into **spout**.

adrilo:
@nimmneun Nobody started working on this enhancement. If you already built something, feel free to share it. Either share a link to your commit or create a pull request.
It's always good to get inspiration from others :)

nimmneun:
[...]

PR #866, opened 2022-02-04
Enable est column width calculation

xwiz:
This is a first step to resolving https://github.com/box/spout/issues/129

This approach primarily computes the width size by using an estimate created

After the PR, the library may be used as follows (sample outputs are attached below):

```
<?php

use Box\Spout\Writer\Common\Creator\WriterEntityFactory;
use Box\Spout\Common\Type;

require __DIR__ . '/vendor/autoload.php';

function generateRows()
{
    $data = [];
    $headers = [];
    for($i=0;$i<10;$i++) {
        $headers[] = "COL-".generateRandomString(5);
    }
    
    $data = [
        $headers
    ];
    
    for($i=0; $i< 10; $i++) {
        $row = [];
        $rand = mt_rand(1, 30);
        foreach ($headers as $header) {
            $randLen = mt_rand(1, $rand);
            $row[] = generateRandomString($randLen);
        }
        $data[] = $row;
    }
    return $data;
}

function generateRandomString($length = 10) {
    $characters = 'aaaaeeeeiiiiiiiioooouuuubcdefghijklmnopqrstuvwxyz';
    $charactersLength = strlen($characters);
    $randomString = '';
    for ($i = 0; $i < $length; $i++) {
        $randomString .= $characters[rand(0, $charactersLength - 1)];
    }
    return $randomString;
}

$data = generateRows();

$writer = WriterEntityFactory::createWriter(Type::ODS);
$writer->setWidthCalculation(1);
$writer->openToFile('spouter.ods');

foreach ($data as $row) {
    $writer->addRow(WriterEntityFactory::createRowFromArray($row));
}

$writer->close();
[spouter.xlsx](https://github.com/box/spout/files/8006663/spouter.xlsx)
[spouter.ods](https://github.com/box/spout/files/8006665/spouter.ods)


?>
```

### DIFF
--- a/src/Spout/Writer/Common/Entity/Worksheet.php
+++ b/src/Spout/Writer/Common/Entity/Worksheet.php
@@ -22,6 +22,21 @@ class Worksheet
 
     /** @var int Index of the last written row */
     private $lastWrittenRowIndex;
+    
+    /** @var array Array of the column widths */
+    protected $columnWidths;
+    
+    /** @var int Width calculation style */
+    protected $widthCalcuationStyle;
+    
+    /** @var int Fixed sheet width for fixed width calculation style */
+    protected $fixedSheetWidth;
+
+    public const W_FULL = 1;
+    public const W_FIXED = 2;
+    public const W_NONE = 0;
+    public const DEFAULT_COL_WIDTH = 30;
+    public const DEFAULT_FIXED_WIDTH = 1068;
 
     /**
      * Worksheet constructor.
@@ -36,6 +51,8 @@ class Worksheet
         $this->externalSheet = $externalSheet;
         $this->maxNumColumns = 0;
         $this->lastWrittenRowIndex = 0;
+        $this->columnWidths = [];
+        $this->widthCalcuationStyle = 0;
     }
 
     /**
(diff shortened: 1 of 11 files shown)


xwiz:
@adrilo I noticed two tests are failing and it seems the test has an error... The two particular resource files involved are generated from a unix environment instead of the current environment.. I think those two files should be dynamically generated to ensure they are compatible with the particular environment..

Meanwhile I've attached further examples of using fixed width style calculation below..

```
<?php

use Box\Spout\Common\Type;
use Box\Spout\Writer\Common\Creator\Style\StyleBuilder;
use Box\Spout\Writer\Common\Creator\WriterEntityFactory;

require __DIR__ . '/vendor/autoload.php';

function generateRows()
{
    $data = [];
    $headers = [];
    for($i=0;$i<10;$i++) {
        $n = $i+1;
        $headers[] = "COL-$n";
    }
    
    $data = [
        $headers
    ];
    
    for($i=0; $i< 10; $i++) {
        $row = [];
        $rand = mt_rand(1, 30);
        foreach ($headers as $header) {
            $randLen = mt_rand(1, $rand);
            $row[] = generateRandomString($randLen);
        }
        $data[] = $row;
    }
    return $data;
}

function generateRandomString($length = 10) {
    $characters = 'aaaaeeeeiiiiiiiioooouuuubcdefghijklmnopqrstuvwxyz';
    $charactersLength = strlen($characters);
    $randomString = '';
    for ($i = 0; $i < $length; $i++) {
        $randomString .= $characters[rand(0, $charactersLength - 1)];
    }
    return $randomString;
}

$data = generateRows();
$defaultStyle = (new StyleBuilder())
    ->setFontSize(36)
    ->build();
$boldtyle = (new StyleBuilder())
    ->setFontSize(42)
    ->setFontBold()
    ->build();
//$writer = WriterEntityFactory::createWriter(Type::ODS);
$writer = WriterEntityFactory::createWriter(Type::XLSX);
$writer->setWidthCalculation(2)->setDefaultRowStyle($defaultStyle)->openToFile('spoutw.xlsx');
//$writer->openToFile('spouter.xlsx');

foreach ($data as $i => $row) {
    if ($i == 0) {
        $writer->addRow(WriterEntityFactory::createRowFromArray($row, $boldtyle));
    } else {
        $writer->addRow(WriterEntityFactory::createRowFromArray($row));
    }
}

$writer->close();
``
[spoutw.xlsx](https://github.com/box/spout/files/8007782/spoutw.xlsx)
[spoutw.ods](https://github.com/box/spout/files/8007783/spoutw.ods)
`

File: src/Spout/Writer/Common/Helper/AppendHelper.php [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * This function will truncate from specified position
+     * Write data to be inserted and re-append the truncated data
+     *  
+     * @param $fp Pointer to file only
+     * @param $pos Position to insert
+     * @param $content Content to insert
+     */
+    public static function insertToFile($fp, $pos, $content)
```
adrilo, line 21: This can be very costly to rewrite the contents, especially with large spreadsheet

src/Spout/Writer/Common/Helper/AppendHelper.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Box\Spout\Writer\Common\Helper;
+
+class AppendHelper {
+
+    /**
+     * Instead of seeking and re-writing from position, a better hack might be to write dummy empty data
+     * Enough to take care of any length, then carefully overwrite
```
adrilo, line 9: Yep, that's the strategy I've used elsewhere

src/Spout/Writer/Common/Entity/Worksheet.php
```diff
@@ -78,6 +95,79 @@ public function getMaxNumColumns()
...
+     * Automatically calculates and sets the max column width for the specified cell
+     * @param Cell $cell The cell
+     * @param Style $style Row/Cell style
+     * @param int $zeroBasedIndex of cell
+     * @return void
+     */
+    public function autoSetWidth($cell, $style, $zeroBasedIndex)
+    {
+        $size = 1 + strlen($cell->getValue());//ensure we have at least 1 space
+        $size *= $style->isFontBold() ? 1.2 : 1.0;
```
adrilo, line 145: How did you come up with the 1.2 ratio?

src/Spout/Writer/Common/Entity/Worksheet.php
```diff
@@ -78,6 +95,79 @@ public function getMaxNumColumns()
...
+    /**
+     * Automatically calculates and sets the max column width for the specified cell
+     * @param Cell $cell The cell
+     * @param Style $style Row/Cell style
+     * @param int $zeroBasedIndex of cell
+     * @return void
+     */
+    public function autoSetWidth($cell, $style, $zeroBasedIndex)
+    {
+        $size = 1 + strlen($cell->getValue());//ensure we have at least 1 space
```
adrilo, line 144: This should be `mb_strlen` to take into account all unicode characters.
I have a concern about this approach: not all characters have the same width: `i` is much narrower than `g` for instance. Errors will compound, so for some data, the cell width will be really off...

Also we need to make sure that it works for values like : `トラベル`

src/Spout/Writer/Common/Entity/Worksheet.php
xwiz: It's actually meant to be 700/400.. but the 1.2 seemed to fit better.. and that may be because tinier letters like i like you rightly pointed out occupy less space..

src/Spout/Writer/Common/Entity/Worksheet.php
xwiz: 🤕 will rectify

src/Spout/Writer/Common/Helper/AppendHelper.php
xwiz: True, I will try both approaches and do a benchmark today. It didn't seem to matter for small files.

xwiz: @adrilo So I ran some tests and this is not actually a bad solution at all... (performance differences range from 3% to 7% and seems to get better for larger amount of rows).. (Only tested 100,000, 500,000 and 1 million rows though)

The difference are quite negligible most likely because the file is still in memory the whole time and the only costly operation here is really calling `stream_get_contents`.

image
image

Issue #829, opened 2021-09-15
Default style with no text wrap is ignored


fwwarr:
Automatic text wrapping due to the presence of new lines in a cell should only be applied if `shouldWrapText` has not been set on the default style. Indeed the `StyleManager` does check `!$cellStyle->hasSetWrapText()` before applying automatic text wrapping, but the following sample shows the opposite behavior:
```
$defaultStyle = (new StyleBuilder())->setShouldWrapText(false)->build();
$writer = WriterEntityFactory::createXLSXWriter();
$writer->setDefaultRowStyle($defaultStyle)->openToFile('testNoWrap.xlsx');
$writer->addRow(WriterEntityFactory::createRowFromArray(["test\n"]));
$writer->close();
```
This seems to be because `StyleMerger::mergeCellProperties` only calls `setShouldWrapText` on the style to update if the `$baseStyle` has `shouldWrapText` true, whereas it probably should call it whenever `hasSetWrapText` is true. 

Current code:
```
        if (!$style->hasSetWrapText() && $baseStyle->shouldWrapText()) {
            $styleToUpdate->setShouldWrapText();
        }
```
Possible fix:
```
        if (!$style->hasSetWrapText() && $baseStyle->hasSetWrapText()) {
            $styleToUpdate->setShouldWrapText($baseStyle->shouldWrapText());
        }
```

Slamdunk:
Fixed in https://github.com/openspout/openspout/pull/63

Issue #737, opened 2020-04-21
Style setting for vertical alignment missing


sclause2412:
Text vertical alignment within cells is missing.

Slamdunk:
Released in [`openspout:v4.2.0`](https://github.com/openspout/openspout/releases/tag/v4.2.0)

PR #869, opened 2022-02-10
Add setCellVerticalAlignment() & allow setShouldWrapText() to be explicitly set as false

jonnott:
Fixes issues:
https://github.com/box/spout/issues/829
https://github.com/box/spout/issues/737

- add Common\Entity\Style\CellVerticalAlignment
- duplicate get/set/hasSet/shouldApply methods for CellAlignment in Common\Entity\Style\Style for CellVerticalAlignment instead, plus corresponding properties
- add setCellVerticalAlignment method to Common\Creator\Style\StyleBuilder
- add vertical alignment to StyleMerger:: mergeCellProperties()
- adjust wrapText logic in mergeCellProperties() to fix issue https://github.com/box/spout/issues/829
- apply vertical cell styling for both XLSX and ODS, via corresponding StyleManager classes
- transform vertical alignment  ‘center’ to ‘middle’ for ODS
- fix logic around wrapText such that the choice whether to include wrapping styles depends on hasSetWrapText() being true, and then use shouldWrapText() thereafter to either set wrapping or no wrapping (for XLSX, wrapText=“1” or wrapText=“0”, for ODS, wrap-option=wrap or wrap-option=no-wrap). previously there was no way to set wrapping to be OFF, only to set it to be ON.
- add new tests to StyleBuilderTest for vertical alignment
- add vertical alignment to documentation.md


src/Spout/Writer/ODS/Manager/Style/StyleManager.php
```diff
@@ -319,6 +336,22 @@ private function transformCellAlignment($cellAlignment)
...
+     * Spec uses 'middle' rather than 'center'
+     * http://docs.oasis-open.org/office/v1.2/os/OpenDocument-v1.2-os-part1.html#__RefHeading__1420236_253892949
+     *
+     * @param string $cellAlignment
+     *
+     * @return string
+     */
+    private function transformCellVerticalAlignment($cellVerticalAlignment)
+    {
+        switch ($cellVerticalAlignment) {
```
adrilo, line 349:
```
return ($cellVerticalAlignment === CellVerticalAlignment::CENTER)
    ? 'middle'
    :  $cellVerticalAlignment;
```

src/Spout/Common/Entity/Style/CellVerticalAlignment.php
```diff
@@ -0,0 +1,34 @@
...
+ * Class CellVerticalAlignment
+ * This class provides constants to work with text vertical alignment.
+ */
+abstract class CellVerticalAlignment
+{
+    public const BOTTOM = 'bottom';
+    public const CENTER = 'center';
+    public const DISTRIBUTED = 'distributed';
+    public const JUSTIFY = 'justify';
+    public const TOP = 'top';
```
adrilo, line 15: According to the ODS spec, here are the valid values:
- auto
- baseline
- bottom
- middle
- top

You have a special transform for "center" (=> "middle"), but what about the others (distributed/justify)?

adrilo (commented):
Thanks @jonnott for your contribution! It looks good overall. I just left 2 minor comments that should be addressed before I can merge.

jonnott:
> Thanks @jonnott for your contribution! It looks good overall. I just left 2 minor comments that should be addressed before I can merge.

Great, I will pop in an extra commit to fix the ODS translations later. I was struggling to find the exact spec details!

src/Spout/Writer/ODS/Manager/Style/StyleManager.php
jonnott: I've adjusted this code as you've suggested.

src/Spout/Common/Entity/Style/CellVerticalAlignment.php
jonnott: I think as there are a completely different set of valid values for vertical alignment for XLSX and ODS respectively, all we can easily do here is allow all of those values to be valid, but still with the transformation of 'center' to 'middle' for ODS, so this is what I've done. Unless you have another suggestion?

jonnott:
> Thanks @jonnott for your contribution! It looks good overall. I just left 2 minor comments that should be addressed before I can merge.

Hi @adrilo, I've force-pushed an amended version of my commit to this PR with the changes as discussed above. Please let me know your thoughts..

jonnott:
> I've force-pushed an amended version of my commit to this PR with the changes as discussed above

@adrilo are we able to merge now?

jonnott:
@adrilo Do you see any outstanding issues with this still preventing a merge?

jonnott:
Anyone know if there's any likelihood that this will get merged any time soon? Or do I need to submit this PR to openspout/openspout and look to use that in my project instead?

@adrilo

Slamdunk:
Merged in https://github.com/openspout/openspout/pull/63 tnx

Issue #846, opened 2022-01-06
Deprecated warnings for incompatible return types in PHP 8.1


fabacino:
Since PHP 8.1 most methods of internal classes require a compatible return type when overriding them (see [Return Type Compatibility with Internal Classes](https://www.php.net/manual/en/migration81.incompatible.php)).

This results in deprecated warnings for the following classes:

* `Box\Spout\Reader\CSV\RowIterator`
* `Box\Spout\Reader\CSV\SheetIterator`
* `Box\Spout\Reader\ODS\RowIterator`
* `Box\Spout\Reader\ODS\SheetIterator`
* `Box\Spout\Reader\Wrapper\XMLReader`
* `Box\Spout\Reader\XLSX\RowIterator`
* `Box\Spout\Reader\XLSX\SheetIterator`

Sample output for `Box\Spout\Reader\CSV\RowIterator`:

```sh
❯ vendor/bin/phpunit | grep -i "deprecated: return type" | grep -i 'csv\\rowiterator'

Deprecated: Return type of Box\Spout\Reader\CSV\RowIterator::current() should either be compatible with Iterator::current(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in src/Spout/Reader/CSV/RowIterator.php on line 227
Deprecated: Return type of Box\Spout\Reader\CSV\RowIterator::next() should either be compatible with Iterator::next(): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in src/Spout/Reader/CSV/RowIterator.php on line 129
Deprecated: Return type of Box\Spout\Reader\CSV\RowIterator::key() should either be compatible with Iterator::key(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in src/Spout/Reader/CSV/RowIterator.php on line 238
Deprecated: Return type of Box\Spout\Reader\CSV\RowIterator::valid() should either be compatible with Iterator::valid(): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in src/Spout/Reader/CSV/RowIterator.php on line 117
Deprecated: Return type of Box\Spout\Reader\CSV\RowIterator::rewind() should either be compatible with Iterator::rewind(): void, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in src/Spout/Reader/CSV/RowIterator.php on line 87
```

[...]

PR #874, opened 2022-03-15
Fix incompatible return types for PHP 8.1

fabacino:
Some return types have already been added here https://github.com/box/spout/commit/e75f6f73012b81fd5fee6107d0af9e86c458448e to get rid of the deprecated warnings in PHP 8.1.
This PR should fix the remaining ones.

Fixes https://github.com/box/spout/issues/846


Slamdunk:
This is a BC break; but has already been fixed in [`openspout/openspout:v3.7.1`](https://github.com/openspout/openspout/releases/tag/v3.7.1) by adding `#[\ReturnTypeWillChange]`

PR #891, opened 2022-05-26
Add "archived project" statement

adrilo:
Because all good things must come to an end... 😢

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,24 @@
 # Spout
 
 [![Latest Stable Version](https://poser.pugx.org/box/spout/v/stable)](https://packagist.org/packages/box/spout)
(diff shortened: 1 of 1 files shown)


Slamdunk:
Hi @adrilo , any chance you could add a link to [`openspout`](https://github.com/openspout/openspout)?

You can warn that the two projects are owned by different people and `Box` doesn't take any responsibility upon `openspout`, but it would help a lot newcomers, especially for upcoming releases of PHP which I'm intended to maintain for a long time.

It would be appreciated very much :pray:

adrilo:
@Slamdunk Unfortunately, I'm not allowed to do that... sorry!

